PR #5, opened 2023-04-19
Generate subprojects

dave-connors-3:
This updates the `split` command to take a prompt for a project name, a selector(s), and the resulting resources into a simple dictionary for later use. 

Not at all married to the design of any of it, mostly just wanted to prove out the use of the Python API to do a `dbt ls` to get the components of a subproject. 

image

### DIFF
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -22,8 +22,9 @@ class LocalDbtProject:
         self.manifest = self.get_project_manifest()
         self.subprojects = []
     
-    def path_to_project(self) -> str:
-        return os.getcwd() + '/' + self.relative_path_to_project
+    @property
+    def path(self) -> os.Pathlike:
+        return os.path.join(os.getcwd(), self.relative_path_to_project)
 
     def dbt_operation(self, operation: List[str]) -> dbtRunnerResult:
         start_wd = os.getcwd()
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -38,7 +38,7 @@ class LocalDbtProject:
             return manifest_result.result
         
     def get_subproject_resources(self, subproject_selector: str) -> List[str]:
-        ls_results = self.dbt_operation(["--quiet", "ls", "-s", subproject_selector])
+        ls_results = self.dbt_operation(["--log-level", "none", "ls", "-s", subproject_selector])
         if ls_results.success:
             return ls_results.result
         
@@ -105,7 +105,7 @@ class LocalDbtProject:
         """
         Returns true if this project depends on the other project as a package or via shared metadata
         """
-        return self.installs(other) or self.shared_source_metadata(other)
+        return self.installs(other) or self.shares_source_metadata(other)
 
 
 
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -20,7 +20,7 @@ class LocalDbtProject:
     def __init__(self, relative_path_to_project: str) -> None:
         self.relative_path_to_project = relative_path_to_project
         self.manifest = self.get_project_manifest()
-        self.subproject_selectors = []
+        self.subprojects = []
     
     def path_to_project(self) -> str:
         return os.getcwd() + '/' + self.relative_path_to_project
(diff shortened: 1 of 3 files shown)


dbt_meshify/dbt_projects.py
```diff
@@ -19,17 +20,33 @@ class LocalDbtProject:
...
+
+    def dbt_operation(self, operation: List[str]) -> dbtRunnerResult:
         start_wd = os.getcwd()
         os.chdir(self.path_to_project())
-        result = dbt_runner.invoke(["--log-level", "none", "parse"])
+        result = dbt_runner.invoke(operation)
         os.chdir(start_wd)
-        if result.success:
-            return result.result
+        return result
```
nicholasyager, line 33: I love that you added this method for interacting with the dbtRunner. This makes it _waaaaaaay_ easier to keep track of directories and paths.

dbt_meshify/dbt_projects.py
```diff
@@ -19,17 +20,33 @@ class LocalDbtProject:
     def __init__(self, relative_path_to_project: str) -> None:
         self.relative_path_to_project = relative_path_to_project
         self.manifest = self.get_project_manifest()
+        self.subprojects = []
     
     def path_to_project(self) -> str:
         return os.getcwd() + '/' + self.relative_path_to_project
```
nicholasyager, line 26: Let's consider turning this into a property. Along the way, we should leverage `os.Pathlike` objects instead of strings for better cross-platform support.

```suggestion
    @property
    def path(self) -> os.Pathlike:
        return os.path.join(os.getcwd(), self.relative_path_to_project)
```

Then, in future calls to this property would just be

```python
self.path
```

or 

```python
project.path
```

instead of calling the `path_to_project()` method

dbt_meshify/dbt_projects.py
```diff
@@ -19,17 +20,33 @@ class LocalDbtProject:
...
+    
+    def get_project_manifest(self) -> Manifest:
+        manifest_result = self.dbt_operation(["--quiet", "parse"])
+        if manifest_result.success:
+            return manifest_result.result
+        
+    def get_subproject_resources(self, subproject_selector: str) -> List[str]:
+        ls_results = self.dbt_operation(["--log-level", "none", "ls", "-s", subproject_selector])
+        if ls_results.success:
+            return ls_results.result
```
nicholasyager, line 43: Non-blocking question: Do we want to raise exceptions returned by the dbtRunner?

nicholasyager (approved):
@dave-connors-3 This is a great start! I left a few python-specific bits of feedback re: properties and exception handling.

**Longer Term Thoughts:**
- I believe there's an opportunity to leverage LocalDbtProject as an abstraction for all dbt projects. If we take this course, I could see `subprojects` in a `LocalDbtProject` themselves being `LocalDbtProject`s.
- It's generally considered good form to create an internal interface class when working with third-party clients. In this case, it _might_ make sense to create a `dbt-meshify`-specific `DbtRunner` class which encapsulates the work that the `dbt-core`.`dbtRunner` performs. In the distant future, what if the runner was actually `dbt-server` 😀

dbt_meshify/dbt_projects.py
dave-connors-3: i would imagine so! not positive how to do that

dave-connors-3:
@nicholasyager totally aligned with the subprojects being `LocalDbtProjects`! I was noodling on that but wasn't sure how to properly do that, given that subprojects don't exist yet! would welcome some code on the right way to handle that abstraction. 

re: a meshify abstraction around the dbtRunner, I am all for it! i imagine that's a separate class that effectively houses the `dbt_operation` method (among other things)?

dbt_meshify/dbt_projects.py
nicholasyager: I think something like this would work:

```python
    def get_subproject_resources(self, subproject_selector: str) -> List[str]:
        ls_results = self.dbt_operation(["--log-level", "none", "ls", "-s", subproject_selector])

        if not ls_results.success:
            raise ls_results.exception

        return ls_results.result
```

Given that this would be common code, there's a good argument for having exception handling performed within the `dbt_operation` method.

nicholasyager:
>re: a meshify abstraction around the dbtRunner, I am all for it! i imagine that's a separate class that effectively houses the dbt_operation method (among other things)?

That's what I'm thinking, too! Definitely work for very near future.

> @nicholasyager totally aligned with the subprojects being `LocalDbtProjects`! I was noodling on that but wasn't sure how to properly do that, given that subprojects don't exist yet! would welcome some code on the right way to handle that abstraction.

Roger that! I'll touch on this in my discussion item re: class design.

PR #25, opened 2023-04-26
Add contract to models

dave-connors-3:
This PR adds a bunch!

### New Classes for File reading and writing

This PR has a few simple classes in `dbt_files.py` to read and write varying filetypes in a dbt project. Currently, it's YML, and everything else 😅 but should be pretty easy to carve out other filetypes (md? json? py?) with this structure

### A New class to actually meshify some stuff

`dbt_meshify` has an early class to edit yml files to add meshify constructs. Right now, it's just one single method that takes in yml from a model's exsiting yml entry, and uses the catalog to add data_types, and adds the `contract` config to the model yml as well. 

### Tests!

That meshify yml editing class also has tests! what a world!


gwenwindflower (approved):

dave-connors-3:
(yolo)

Issue #27, opened 2023-04-26
update file writing class to use pathlib


dave-connors-3:
should make some of the classes I made in the file writing lib a lot simpler

PR #28, opened 2023-04-28
add standalone add contract command

dave-connors-3:
This PR exposes the logic from #25 as a standalone command for adding contracts to resources. 

I think the edits here to the new command `contract` mirror some changes we can and should make to the other commands, namely, assume that the project is in the cwd unless otherwise specified by a flag. this gives us some flexibility, and makes the commands look and feel a bit more like dbt itself. 

I do not have any idea how to write a good integration test here :) Would really love some help on that!


dbt_meshify/main.py
```diff
@@ -51,3 +51,26 @@ def split():
         holder.register_project(subproject)
 
     print(holder.project_map())
+
+
+@cli.command(name="contract")
```
nicholasyager, line 56: Non-blocking feedback: I wonder if this command should be named more descriptively for what it does. For instance, maybe `generate_contracts` or some other phrasing that indicates that contracts are being created?

dbt_meshify/dbt_projects.py
```diff
@@ -184,6 +168,24 @@ def split(
...
+    def add_model_contract(self, unique_id: str) -> None:
+        """Adds a model contract to the model's yaml"""
+        
+        # get the patch path for the model
+        node = self.get_manifest_node(unique_id)
+        yml_path = node.patch_path.split("://")[1] if node.patch_path else None
+        # if the model doesn't have a patch path, create a new yml file in the models directory
+        # TODO - should we check if there's a model yml file in the models directory and append to it?
+        if not yml_path:
+            yml_path = "/".join(node.original_file_path.split(".")[0].split("/")[:-1]) + "_models.yml"
```
nicholasyager, line 180: Very clever way to parse the file name without the extension.

Another approach for this would be to leverage the `pathlib` library's stem property.

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem

dbt_meshify/main.py
```diff
@@ -51,3 +51,26 @@ def split():
...
+    )
+@click.option(
+    "--project-path",
+    default="."
+)
+def contract(select, exclude, project_path):
+    path = Path(project_path).expanduser().resolve()
+    project = DbtProject.from_directory(path)
+    resources = list(project.select_resources(select=select, exclude=exclude, output_key="unique_id"))
+    models = [json.loads(resource)["unique_id"] for resource in resources if json.loads(resource)["unique_id"].startswith("model")]
```
nicholasyager, line 73: This seems a little cumbersome. Perhaps parsing json results into `List[Dict]` should be a responsibility of `select_resources`?

Also, once you have an iterable (like List or Set) you can use the `filter()` function with a `lambda` to filter the list down to values starting with `"model"`.

Something like this, assuming resources is `List[Dict[str, str]]`.
```python
models = filter(lambda x: x['unique_id'].startswith('model'), resources)
```

dbt_meshify/dbt_meshify.py
```diff
@@ -33,10 +35,12 @@ def add_model_contract_to_yml(self, full_yml_dict: Dict[str, str], model_catalog
...
+        model_ordered_dict.update(model_yml)
+        model_ordered_dict = {k: v for k, v in model_ordered_dict.items() if v is not None}
         if not full_yml_dict.get("models"):
-            full_yml_dict.update({"models": [model_yml]})
+            full_yml_dict.update({"models": [model_ordered_dict]})
         else:
             for i, entry in enumerate(full_yml_dict.get("models", [])):
                 if entry["name"] == model_name:
-                    full_yml_dict["models"][i] = model_yml
+                    full_yml_dict["models"][i] = model_ordered_dict
```
nicholasyager, line 45: I suspect we'll eventually want to manage an internal dictionary structure for managing models. This way, we will be able to pass in the name of the model and directly update the values without needing to iterate through a list.

nicholasyager (approved):
@dave-connors-3 On the whole, this looks good. I added some thoughts about the current implementation, but all the feedback should be treated like ideas for future refinement rather than requested changes.

dbt_meshify/main.py
dave-connors-3: i like that idea! i was optimizing for verbs, and brevity, thinking that some other commands could be `publicize`, `group` and the like, but there are likely limits to that approach

dbt_meshify/dbt_projects.py
dave-connors-3: totally aligned, i think there's a lot of cleanup to be done on the file manager class with pathlib -- opened #27 to reflect that!

dbt_meshify/main.py
dave-connors-3: agreed here as well! i was thinking the `output_key` argument would only ever be one thing (for now at least), so the filtering and reducing can be housed a layer down for sure -- i don't love the `json.loads()` here either, but i think the output from the ls is a little wonky when it comes back as json!

dbt_meshify/dbt_meshify.py
dave-connors-3: agreed here -- what might that look like roughly speaking?

nicholasyager: One solution would be to have a variable that stores the dictionary, and to grab the dictionary items for the models list just before returning. Maybe something like:

```python

models = { model['name']: model for model in full_yml_dict['models'] }

# ... stuff already in the method ...

# Update the model entry
model_name = model_ordered_dict['name']
models[model_name] = model_ordered_dict

# ... anything else needed ...

# return the yml
full_yml_dict['models'] = list(models.values())
return full_yml_dict
```

dbt_meshify/main.py
nicholasyager: Let's roll with it as is, and see how it feels! Definitely not something to bike shed this early 😝

nicholasyager:
@dave-connors-3 w/r/t your comment around integration testing, I believe you can use a Click CliRunner with filesystem isolation to invoke the command with some preconfigured values and verify that the generated yml is appropriate.

Here's a basic overview of what's possible: https://click.palletsprojects.com/en/8.1.x/testing/

dbt_meshify/dbt_meshify.py
dave-connors-3: made this change as well -- think i could see the argument for us maintaining a ModelYml class that handles the allowed keys, data types, and order of keys we want to display, so it could endup like 

```python
models = { model['name']: ModelYml(model) for model in full_yml_dict['models'] }
```

test-projects/source-hack/src_proj_a/.gitignore
```diff
@@ -2,3 +2,5 @@
 target/
 dbt_packages/
 logs/
+**/*.database.db
+**/*.duckdb.wal
```
nicholasyager, line 6: Fantastic call!

dbt_meshify/dbt_projects.py
```diff
@@ -208,6 +190,30 @@ def split(
...
+        """Adds a model contract to the model's yaml"""
+        
+        # get the patch path for the model
+        node = self.get_manifest_node(unique_id)
+        yml_path = Path(node.patch_path.split("://")[1]) if node.patch_path else None
+        original_file_path = Path(node.original_file_path) if node.original_file_path else None
+        # if the model doesn't have a patch path, create a new yml file in the models directory
+        # TODO - should we check if there's a model yml file in the models directory and append to it?
+        if not yml_path:
+            yml_path = original_file_path.parent / "_models.yml"
```
nicholasyager, line 203: Woah! This operator is going to come in handy

dbt_meshify/dbt_meshify.py
```diff
@@ -7,19 +8,19 @@ class DbtMeshYmlEditor:
...
-        # import pdb; pdb.set_trace()
-        if full_yml_dict:
-            model_yml = [entry for entry in full_yml_dict.get("models", {}) if entry["name"] == model_name].pop()
-        else:
-            full_yml_dict = {"models": []}
-            model_yml = {"name": model_name, "columns": [], "config": {}}
+        # set up yml order
+        model_ordered_dict = OrderedDict.fromkeys(["name", "description", "access", "config", "meta","columns"])
+        # parse the yml file into a dictionary with model names as keys
+        models = { model['name']: model for model in full_yml_dict['models'] } if full_yml_dict else {}
```
nicholasyager, line 16: I'm so happy to see this added! Fantastic enhancement 🚀

dbt_meshify/file_manager.py
```diff
@@ -67,14 +67,14 @@ def __init__(self,
 
     def read_file(self, path: os.PathLike) -> Union[Dict[str, str], str]:
         """Returns the yaml for a model in the dbt project's manifest"""
-        if path.endswith('.yml'):
+        if path.suffix == '.yml':
             return self.yml_file_manager.read_file(path)
         else:
             return self.base_file_manager.read_file(path)
```
nicholasyager, line 73: ⚠️ **Slightly-irrelevant feedback**

This is outside the scope of your change, but here's a more succinct way of writing this in the future.

```python
if path.suffix == '.yml':
    return self.yml_file_manager.read_file(path)
return self.base_file_manager.read_file(path)
```

nicholasyager (approved):
@dave-connors-3 your latest rounds of changes definitely improve the ergonomics of this code quite a bit. Thanks for the alterations 🚀

PR #29, opened 2023-04-30
Feature: Create standardized `Mesh` class for cross-project dependency detection

### LINKED ISSUE #9: standardized method/class to detect dependcies between projects
- early version is `LocalDbtProject.overlapping_sources()` which tells you the sources in the downstream project that point to models in the upstream project, but this is not really the way want to do this. should be more generic

nicholasyager:
# Description of Changes

This pull request implements the `Mesh` class, which serves as the primary method for operations that affect the "meshiness" of a set of project, including dependency detection. In the future, this class may also manage logic around group _detection_ and reflowing cross-project resource references.

Currently, this model has a single public method `dependencies()`, which returns a set of `ProjectDependency`s between two `BaseDbtProject`s. Dependency logic now only encompasses source hack detection and package imports. Different dependency types are represented by the `ProjectDependency.type` property, which can currently be `Source` or `Package`.

For example:

```python

# In this scenario, Project A defines `model_a`, and Project B has a source 
# with the same relation name as `model_a`.

mesh = Mesh()
dependencies: Set[ProjectDependency] = mesh.dependencies(project_a, project_b)

print(dependencies)

{
    ProjectDependency(
        upstream: 'model.project_a.model_a',
        downstream: 'source.project_b.project_a.model_a',
        type: ProjectDependecyType.Source
    )
}
```

Along the way, I also overhauled the `test-projects` directory! This includes
- Converting all test projects from postgres to duckdb for _easy_ integration testing.
- Fixing profile and `dbt_project.yml` defects.

And I implemented integration tests for current `Mesh` happy paths.

Resolves #9


dbt_meshify/mesh.py
```diff
@@ -0,0 +1,134 @@
...
+    ) -> Set[ProjectDependency]:
+        """
+        Identify source-hack dependencies between projects.
+
+        Source hacking occurs when Project A defines a model, and a downstream project (Project B)
+        defines a source for the materialization of that same model.
+        """
+
+        relations = self._find_relation_dependencies(
+            source_relations={model.relation_name for model in project.models().values()},
```
dave-connors-3, line 63: would this be the same as `project.model_relation_names`?

dave-connors-3: ah one it `items` one is `values`!

dbt_meshify/mesh.py
```diff
@@ -0,0 +1,134 @@
...
+        backward_dependencies = {
+            ProjectDependency(
+                upstream=other_project.model_relation_names.get(relation),
+                downstream=project.source_relation_names.get(relation),
+                type=ProjectDependencyType.Source,
+            )
+            for relation in backwards_relations
+        }
+
+        return forward_dependencies | backward_dependencies
```
dave-connors-3, line 90: i think this pipe syntax might be problematic for lower python versions?

dave-connors-3 (approved):
really exciting! one small comment, but looks great to m

dbt_meshify/mesh.py
nicholasyager: The `set1 | set2` notation is supported for unions starting in Python 3.5, which is currently EOL. We should be good!

https://docs.python.org/3.5/library/stdtypes.html#frozenset.union

dave-connors-3: beautiful! i thought that was an `or` !

PR #31, opened 2023-05-03
Add model version command

dave-connors-3:
This PR extends and refactors some of the class structure introduced in #28 to better decouple dbt project properties with mesh operations, and adds a `version` command to add/increment model versions in yml and create the appropriate files. I think there's likely some discussion to be had on both the refactoring work here (do the mesh classes make sense?) as well as the assumptions around file creation. 

 I think my python code is always too verbose, so if there are ways to make my logic more succinct would relish that feedback!


dbt_meshify/dbt_meshify.py
```diff
@@ -34,10 +47,127 @@ def add_model_contract_to_yml(self, model_name: str, model_catalog: CatalogTable
...
+        # create the new version file 
+        
+        # if we're incrementing the version, write the new version file with a copy of the code
+        latest_version = node.latest_version if node.latest_version else 0
+        last_version_file_name = f"{node.name}_v{latest_version}.{node.language}"
+        next_version_file_name = f"{defined_in}.{node.language}" if defined_in else f"{node.name}_v{latest_version + 1}.{node.language}"
+        next_version_path = original_file_path.parent / next_version_file_name
+        last_version_path = original_file_path.parent / last_version_file_name
+        
+        # import pdb; pdb.set_trace()
```
nicholasyager, line 160:
```suggestion
```

dbt_meshify/dbt_meshify.py
```diff
@@ -34,10 +47,127 @@ def add_model_contract_to_yml(self, model_name: str, model_catalog: CatalogTable
...
+        if defined_in:
+            version_dict["defined_in"] = defined_in
+        # add the version to the model yml entry
+        versions_list.append(version_dict)
+        # update the latest version in the model yml entry
+        model_yml["versions"] = versions_list
+        model_yml["latest_version"] = latest_version
+
+        processed = self.process_model_yml(model_yml)
+        models[model_name] = processed
```
nicholasyager, line 86: ⚠️ Non-blocking feedback

I'm not quite sure where this ought to go. I know that there's a need for an ergonomic abstraction for working with YML and saving model properties. I suspect we'll be well served by letting this PR ride, and then coming back to redesign this abstraction.

dbt_meshify/dbt_projects.py
```diff
@@ -154,9 +154,10 @@ def __init__(
         super().__init__(manifest, project, catalog, name)
         self.path = path
         self.dbt = dbt
-        self.file_manager = DbtFileManager(read_project_path=path, write_project_path=path)
 
-    def select_resources(self, select: str, exclude: Optional[str] = None, output_key: Optional[str] = None) -> Set[str]:
+    def select_resources(
+        self, select: str, exclude: Optional[str] = None, output_key: Optional[str] = None
+    ) -> Set[str]:
```
nicholasyager, line 160: Hooray for improved formatting!

nicholasyager (approved):

dbt_meshify/dbt_meshify.py
dave-connors-3: do you have early thoughts here? happy to take a crack at some of these refactoring tasks this week

PR #36, opened 2023-05-11
Feature: Support for creating groups and automatically assigning models based on node selection

### LINKED ISSUE #21: Identify candidates for `groups` and `private` access
https://docs.getdbt.com/docs/collaborate/govern/model-access

As the maintainer of a large & complex project, I'd be interested in a tool that can help me identify / recommend:
- potential `groups` for a set of currently ungrouped models
- ungrouped models that are good candidates for existing `groups`
- protected models that could be switched to `access: private`

### Proposed criteria

Many dbt developers follow [our recommended practices](https://docs.getdbt.com/guides/best-practices/how-we-structure/1-guide-overview) around project structure, and so they're already in the habit of grouping together related models in subdirectories. I expect that this will be the way many teams start using groups, by adding a `group` config in `dbt_project.yml` for all models in that subdirectory, and defining an owner for those models. This could be a naive starting point for us too.

At the cleverer end of the spectrum, we could cook up a DAG simplification / [edge contraction](https://en.wikipedia.org/wiki/Edge_contraction) algorithm, where all removed vertices should be private models in the same group.

Given:
```
model_a --> [one or more models] --> model_b
```
Where the only children of those models is `model_b`, they should be switched to `private` models within the same group as `model_b`.

I'm sure there are even cleverer approaches we could take here!

nicholasyager:
Resolves #21 

# Description of Changes

This PR has a combination of minor and major changes.

## Major Changes
- Addition of the `create-group` CLI command, which allows for adding nodes to a Group based on NodeSelection syntax.
- Creating a new `ResourceGrouper` class it is capable of creating new `Groups` and `Owners` for a project. This includes making some educated guesses about what access types should be applied to different nodes in the selection.
- Updates to `DbtMeshYmlEditors` to edit `group` values, and to edit `models` to include `group` and `access` values.

## Minor Changes
- Renamed `Mesh` to `Linker` for more generic naming. Since this project is `dbt-meshify`, it's weird to call anything `mesh`.
- Moved files around to make it easier to know what files do. Storage related files are now in `storage/`, including `yml_editors` and `file_manager`.
- Added tests for various parts of the create-group command and the underlying tech.

# Access Type Inference

The `ResourceGrouper` class uses the following logic for inferring between `public` and `private` access.
1. If the node has no downstream children, then set access to `public`. The rationale is that these nodes may be report models or marts.
2. If the node has downstream children outside the selected node set, then set access to `public`. We cannot set this to private, since there are already references from beyond the group.
3. If the node has downstream children, all of which are inside the selected node set, then we can assume that the model is internal to the group. Set the node's access to `private`.

```mermaid
flowchart LR

    classDef source fill:#60b826, stroke:#60b826, color:#fff
    classDef model fill:#0094b3, stroke:#0094b3, color:#fff
    classDef newModel fill:#b16fb1, stroke:#b16fb1, color:#fff
    classDef privateModel fill:#fff, stroke:#0094b3, color:#0094b3

    subgraph group
        source:::source
        staging:::privateModel
        intermediate:::privateModel
        mart:::model
        mart_2:::model
    end

    source --> staging --> intermediate --> mart
    intermediate --> mart_2 --> reporting:::model

```


tests/integration/test_create_group_command.py
```diff
@@ -0,0 +1,126 @@
...
+        create_group,
+        [
+            "test_group",
+            "--select",
+            "shared_model",
+            "--project-path",
+            proj_path_string,
+            "--owner",
+            "name",
+            "Shaina Fake",
```
dave-connors-3, line 55: hell yeah

dbt_meshify/dbt_projects.py
```diff
@@ -43,6 +44,32 @@ def __init__(
...
+        self._graph = None
+
+        self._changes: Dict[str, str] = {}
+
+
+    @staticmethod
+    def _load_graph(manifest: Manifest) -> Graph:
+        """Generate a dbt Graph using a project manifest and the internal dbt Compiler and Linker."""
+
+        from dbt.compilation import Compiler
```
dave-connors-3, line 56: is having imports within a method a stylistic choice, or are there performance benefits (or neither)?

dbt_meshify/dbt_projects.py
```diff
@@ -121,6 +148,13 @@ def get_manifest_node(self, unique_id: str) -> ManifestNode:
         """Returns the catalog entry for a model in the dbt project's catalog"""
         return self.manifest.nodes.get(unique_id, {})
 
+    def select_resources(
```
dave-connors-3, line 151: Is this necessary to include in this class? i see it's overridden in the inherited class, i'm just curious about the utility to also include it here.

dbt_meshify/dbt_projects.py
```diff
@@ -156,7 +190,9 @@ def __init__(
         self.dbt = dbt
         self.file_manager = DbtFileManager(read_project_path=path, write_project_path=path)
 
-    def select_resources(self, select: str, exclude: Optional[str] = None, output_key: Optional[str] = None) -> Set[str]:
+    def select_resources(
```
dave-connors-3, line 193: aside: would love to also have a) black as part of the requirements and b) some pre-commit stuff setup so that we are all using it consistently!

dbt_meshify/utilities/grouper.py
```diff
@@ -0,0 +1,131 @@
...
+        exclude: Optional[str] = None,
+    ) -> Tuple[Group, Dict[str, AccessType]]:
+        """Generate the ResourceGroup that we want to apply to the project."""
+
+        group = Group(
+            name=name,
+            owner=owner,
+            package_name=self.project.name,
+            original_file_path=os.path.relpath(path, self.project.path),
+            # TODO: This seems a bit yucky. How does dbt-core generate unique_ids?
```
dave-connors-3, line 74: [appears to be just how you did it](https://github.com/dbt-labs/dbt-core/blob/8f998c218eb8ec37bc3db99ac93148be755f332c/core/dbt/parser/base.py#L52)

dbt_meshify/storage/yaml_editors.py
```diff
@@ -0,0 +1,102 @@
...
+class DbtMeshYmlEditor:
+    """
+    Class to operate on the contents of a dbt project's dbt_project.yml file
+    to add the dbt-core concepts specific to the dbt linker
+    """
+
+    @staticmethod
+    def add_group_to_yml(group: Group, full_yml_dict: Dict[str, Any]):
+        """Add a group to a yml file"""
+        if full_yml_dict is None:
```
dave-connors-3, line 18: do you think we can a) tighten up this logic and/or b) abstract this into methods at all? i think some of the boilerplate read-and-set-up-a-dict and filter-and-write-out the yml stuff might be really similar across methods

dave-connors-3 (approved):
looks really good -- a couple cleanup and future looking thoughts, but i love this feature a whole lot. let's chat about sequencing this with my other PR -- this one is higher prio, but there are definitely some overlaps in the class structure we'll want to coordinate on

dbt_meshify/dbt_projects.py
nicholasyager: I'm doing it more from a performance perspective. Top-of-file imports result in all of the project's dependencies being loaded from disk at the time of interpretation. Normally this is great, but for a CLI-based tool, it can create some lag during invocation. In our case, the `dbt-core` package is a little on the big side, so I'm trying to get into the habit of only importing locally when possible.

dbt_meshify/dbt_projects.py
nicholasyager: Strong agree!

dbt_meshify/storage/yaml_editors.py
nicholasyager: I do! Ideally, this type of low-level list/dict manipulation can be abstracted into some other structure. I'll have to think a little more as to what that ought to be!

dbt_meshify/utilities/grouper.py
nicholasyager: Lol! I guess we're good to go, then 🤣 

```suggestion
```

dbt_meshify/dbt_projects.py
nicholasyager: Nah, this can go in the bin.

dbt_meshify/utilities/grouper.py
```diff
@@ -0,0 +1,130 @@
...
+    def classify_resource_access(cls, graph, nodes):
+        """
+        Identify what access types each node should have. We can make some simplifying assumptions about
+        recommended access for a group.
+
+        For example, interfaces (nodes on the boundary of a subgraph or leaf nodes) should be public,
+        whereas nodes that are not a referenced are safe for a private access level.
+        """
+        boundary_nodes = cls.identify_interface(graph, nodes)
+        print(boundary_nodes)
```
dave-connors-3, line 52: may not need this print anymore!

nicholasyager: That can definitely go!

.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
-oudefault_stages: [push]
```
nicholasyager, line 1: This is an invalid property! I've corrected this accordingly.

nicholasyager:
@dave-connors-3 This has been updated to be compatible with `main`!

PR #37, opened 2023-05-15
add Precommit for isort and black

dave-connors-3:
We should have some standard tooling for code quality! This PR adds 

Packages to project toml:
- `pre-commit`
- `black` (for code formatting)
- `isort` (cleans up imports)
- `mypy`

Files to repo
- `pre-commit-config.yaml` - defines the precommit rules! 

I am commenting out flake8 for now, since it got pretty noisy, but happy to include if we feel it's necessary. 

I applied an initial `poetry run black -S .` here to also standardize the code, so lots of edits here, but should be noncontroversial

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,53 @@
+default_stages: [push]
+default_language_version:
+  python: python3.9
+repos:
+- repo: local
+  hooks:
+  - id: isort
+    stages: [commit,push]
+    name: isort
+    entry: poetry run isort -rc
+    language: system
+    types: [python]
+  - id: black
+    stages: [commit,push]
+    name: black
+    entry: poetry run black -S .
+    language: system
+    types: [python]
+  # - id: mypy
+  #   stages: [commit,push]
+  #   name: mypy
+  #   entry: poetry run mypy --ignore-missing-imports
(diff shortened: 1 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+    stages: [commit,push]
+  - id: check-json
+  - id: check-merge-conflict
+    stages: [commit,push]
+  - id: check-symlinks
+  - id: check-vcs-permalinks
+  - id: check-xml
+  - id: check-yaml
+  - id: debug-statements
+  - id: detect-private-key
```
nicholasyager, line 46: Yiiiiis!

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+  # - id: flake8
+    # stages: [commit,push]
```
nicholasyager, line 48: I see this is commented out. A reasonable alternative compatible w/ Black might be [ruff](https://github.com/charliermarsh/ruff).

nicholasyager (approved):
I am sooooo happy to see this PR come through. My only bit of feedback in on an alternative to flake8, which I saw commented out. Otherwise, Let's goooooo 🚀

PR #41, opened 2023-05-16
Support for `--selector` 

dave-connors-3:
this PR adds support for the `--selector` cli flag we have stued, but doing nothing! simply passes the `--selector` to the `ls` operation under the hood!

Opted not to add any tests, but that may have been unwise!


nicholasyager (approved):
Short and sweet. This is a fantastic change @dave-connors-3!

graciegoheen (approved):

PR #43, opened 2023-05-17
remove duckdbs

dave-connors-3:
This PR removes the duckdb files from test projects, updates their git ignores, and adds a `dbt run` step to the `test_contract_command.py` file before any operation happens to make sure the models exist before the contract command calls the `dbt docs generate` step


nicholasyager (approved):
🚀

graciegoheen (approved):

PR #44, opened 2023-05-23
Code cleanup

dave-connors-3:
This PR is an attempt to simplify some existing code

1. Rework the `file_manager.py` -- these classes may still be a little heavier than we need -- effectively, I have reduced this down to a small wrapper around the `pathlib` read and write operations to handle yml in one go, rather than having two additional classes just to handle this usecase. In reality, these could just be functions, but opted to keep it within a class so we have read from and write to paths for use in the future where we're moving files between projects. 
2. Attempt to simplify the methods in our `dbt_meshify.py` file:

  - pull out common operations into helper functions outside the classes
  - rename some variables for clarity (full_yml_dict --> models_dict or groups_dict most notably) 
  - pull out a couple common filepath related operations into methods for the actual constructor

The methods in `dbt_meshify.py` are still a bit beefy -- would love to think through how we can simplify them further in this PR!

### DIFF
--- a/dbt_meshify/storage/file_manager.py
+++ b/dbt_meshify/storage/file_manager.py
@@ -3,6 +3,7 @@
 
 import os
 from abc import ABC
+from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 import yaml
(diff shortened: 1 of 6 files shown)


nicholasyager (approved):
@dave-connors-3 Thanks for going through and making these changes! I suspect this will make it quite a bit easier to work on file management in the future. 🚀 ✅

PR #51, opened 2023-05-31
add missing package

dave-connors-3:
the workflows needed to also include the package used to render the click help text in the docs. May be worth using poetry in the future


nicholasyager (approved):
@dave-connors-3 Assuming the necessary github_token secret has been added to the repo, this looks great to me!

PR #57, opened 2023-06-07
Add larger length size to prevent wrapping lines

b-per:
The demo showed long YAML lines being wrapped, generating a bigger diff than needed.
This should fix it


dave-connors-3 (approved):
look at this fella go

Issue #59, opened 2023-06-08
Update to `pyproject.toml` dependencies


b-per:
The current file has this section

```
[tool.poetry.dependencies]
python = "^3.9"
dbt-core = "^1.5.0rc1"
click = "^8.1.3"
dbt-postgres = { version = "^1.5.0rc1", optional = true }
dbt-duckdb = "^1.5.0"
black = "^23.3.0"
mypy = "^1.3.0"
isort = "^5.12.0"
pre-commit = "^3.3.1"
ruamel-yaml = "^0.17.31"
mike = "^1.1.2"
```

Shouldn't some of them move to `[tool.poetry.group.dev.dependencies]`:
```
[tool.poetry.group.dev.dependencies]
...
mypy = "^1.3.0"
isort = "^5.12.0"
pre-commit = "^3.3.1"
mike = "^1.1.2"
```

PR #60, opened 2023-06-08
Build: Move development dependencies into `dev` dependency group

nicholasyager:
# Description and Motivation

This PR moves dev dependencies like `isort`, `black`, and `mike` into the `dev` dependency group in `poetry`. Moving forward, you will need to specify the `dev` group when install dependencies locally if you want to develop. 

For example:
```console
poetry install --with=dev
```

I've also updated the dependency installation steps in our GitHub actions accordingly.

Resolves: #59


dave-connors-3 (approved):
🦞

PR #62, opened 2023-06-13
add pypi publish workflow

graciegoheen:
Created a github workflow to publish release of package to pypi

- created a new github environment "PypiProd" (to restrict the reviewers) 
- created a new tag protection rule (to restrict who can create new tags)
- added publish.yml to the repo following [this](https://pgjones.dev/blog/trusted-plublishing-2023/) article

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,8 +1,7 @@
 name: Publish
 on:
-  push:
-    tags:
-      - '*'
+  release:
+    types: [created]
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,44 @@
+name: Publish
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: 3.11
+
+    - name: Install dependencies and build
+      run: |
+        python3 -m pip install --upgrade pip
+        pip install poetry
+        poetry build
+
+    - name: Upload artifact
+      uses: actions/upload-artifact@v3
+      with:
+        path: ./dist
+
+  pypi-publish:
+    needs: ['build']
+    environment: 'PypiProd'
+
+    name: upload release to PyPI
+    runs-on: ubuntu-latest
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+      - uses: actions/download-artifact@v3
+
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages_dir: artifact/
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,9 +11,9 @@ jobs:
       uses: actions/checkout@v3
 
     - name: Set up Python
-      uses: actions/setup-python@v3
+      uses: actions/setup-python@v4
       with:
-        python-version: 3.11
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+      - '*'
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v3
```
emmyoop, line 14:
```suggestion
      uses: actions/setup-python@v4
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+      with:
+        python-version: 3.11
```
emmyoop, line 16: I would be consistent here with the version of python.  Looks like you use 3.9 elsewhere, so I would 3.9 here.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+    needs: ['build']
+    environment: 'PypiProd'
+
+    name: upload release to PyPI
+    runs-on: ubuntu-latest
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+    steps:
+      - uses: actions/download-artifact@v3
```
emmyoop, line 39:
```suggestion
      - name: Download artifacts
        uses: actions/download-artifact@v3
```
I like to name all my steps.  It just makes the output logs a bit easier to read.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
+name: Publish
+on:
+  push:
+    tags:
+      - '*'
```
emmyoop, line 5: Is your plan to release only when you manually push a tag?  The workflow you mentioned in slack published to PyPI after they [create a new release](https://github.com/dpguthrie/dbtc/blob/1e83a4031f4dc89c99ebd32de6d7dce222d5ec11/.github/workflows/publish.yml#L3-L5).  Either is fine, just want to be clear on when this gets triggered.

Core has a more complex design that tags it and then releases it for you to both GitHub and PyPI.

emmyoop (changes requested):
Looks great!  Just a few changes and a question.

.github/workflows/publish.yml
graciegoheen: Screen Shot 2023-06-13 at 12 56 44 PM

I think the plan was to publish to PyPI when we create a new release. I might be confused here because I thought creating a new release _also_ involved creating a new tag on publish.

graciegoheen, line 5:
```suggestion
on:
  release:
    types: [created]
```

.github/workflows/publish.yml
graciegoheen, line 16:
```suggestion
        python-version: 3.9
```

.github/workflows/publish.yml
dave-connors-3: i think we're unlikely to create tags that are not releases, but i would say the release event is more specific and accurate

emmyoop: I prefer being intentional as well.  This way someone can't accidently push a tag that generates a release.

graciegoheen: Gotcha - sounds like release is the way to go then. Thanks for the explanation

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
+name: Publish
+on:
```
graciegoheen, line 2:
```suggestion

on:
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,44 @@
...
+  release:
+    types: [created]
```
graciegoheen, line 4:
```suggestion
    types: [created]

```

emmyoop (approved):
Looks good to me!

PR #63, opened 2023-06-13
Split command

dave-connors-3:
Opening up a PR to collaborate on the `split` command!

Right now, this command:

- accepts the standard args for `select/exclude/selector` to identify which resources should be moved
- registers those resources as a subproject of the original project
- calls the `subproject.initialize()` method. 

The `initialize()` method:
 - iterates over the resources in the subproject
   - for resources represented by a `.sql` , `.py`, or `.csv` file, move that file into the subdirectory, then move the resource's yml _entry_, if necessary
   - for `.yml` defined resources, copy over only the resource yml entry. 
 - writes the subproject's `dbt_project.yml`
 
These changes requires some new `DbtYmlEditor` methods, especially to handle the nuance of source files and make our methods more generic to other resource types that follow similar patterns. Additionally, there's a lot of variable renaming for accuracy's sake, so apologies in advance for lots of edits. 

Things this command does not do, but will need to before shipping:
- declare model access and groups for boundary nodes / leaf nodes of selected resources (question: do we need `group`s at all?)
- edit children of boundary nodes / leaf nodes to use the multi-argument ref function
- write a `dependencies.yml` file to the parent project
- copy necessary macros from parent project to subproject
- tests!
- anything else we can think of!
 
 I would really love some input here on syntax, logic, and design. This is not at all complete, but


dave-connors-3:
@nicholasyager -- this now also moves relevant custom macros and blindly copies the packages.yml to the new subdirectory. There may be a small bug in one of the yml entry moving methods that I am trying to track down, but if you run a 

```zsh
> dbt-meshify split my_project --select +orders
> cd my_project
> dbt deps && dbt compile
```

you _should_ get into a runnable state 🤞 let me know if that is the case! then we can work to actually make this code good which it currently is not

dbt_meshify/dbt_projects.py
```diff
@@ -242,13 +261,51 @@ class DbtSubProject(BaseDbtProject):
...
+        for unique_id in self.resources:
+            resource = self.get_manifest_entry(unique_id)
+            if not resource:
+                continue
+            macros = resource.depends_on.macros
+            project_macros = [
+                macro
+                for macro in macros
+                if hashlib.md5((macro.split(".")[1]).encode()).hexdigest()
+                not in self.parent_project.installed_packages()
```
dave-connors-3, line 305: could probably change this check to `== self.manifest.metadata.project_id`

dbt_meshify/dbt_projects.py
```diff
@@ -285,12 +342,52 @@ def split(
...
+        Writes the dbt_project.yml file for the subproject in the specified subdirectory
+        """
+        contents = self.project.to_dict()
+        # was gettinga  weird serialization error from ruamel on this value
+        # it's been deprecated, so no reason to keep it
+        contents.pop("version")
+        # this one appears in the project yml, but i don't think it should be written
+        contents.pop("query-comment")
+        contents = filter_empty_dict_items(contents)
+        self.file_manager.write_file(Path("dbt_project.yml"), contents)
```
dave-connors-3, line 356: may not want the file manager in this class -- to date, it's only in the `DbtMeshConstructor` class that's operating on resource files.

nicholasyager:
@dave-connors-3 I plan on looking at the code tomorrow evening, but in the meantime I did have a chance to noodle with the command as it's currently implemented. 

My specific invocation, for reference:

```
poetry run dbt-meshify split --select +orders revenue 
```

Here are some thoughts in no particular order:
- Since we're copy `packages.yml`, we could also copy over the `dbt_packages` directory too. This will let the user needing to run `dbt deps` before their first build of the subproject.
- We may want to have some logic in place that redistributes/partitions groups into the subproject. Currently, if a model is part of a group and that model is moved to a new project, it is retaining its old group and compilation fails.
- As a user, it would be useful to have information logged during execution communicating what has been done. After running the `split` command for the first time, I was left thinking, "huh. I wonder if that worked." 😝

dbt_meshify/dbt_projects.py
```diff
@@ -150,9 +154,21 @@ def get_catalog_entry(self, unique_id: str) -> Optional[CatalogTable]:
         """Returns the catalog entry for a model in the dbt project's catalog"""
         return self.catalog.nodes.get(unique_id)
 
-    def get_manifest_node(self, unique_id: str) -> Optional[ManifestNode]:
+    def get_manifest_entry(self, unique_id: str) -> Optional[ManifestNode]:
         """Returns the catalog entry for a model in the dbt project's catalog"""
```
nicholasyager, line 158: I don't quite agree with renaming this method. This specifically is getting a `ManifestNode`, which would indicate that `get_manifest_node` would be an appropriate name.

dbt_meshify/dbt_projects.py
```diff
@@ -242,13 +261,51 @@ class DbtSubProject(BaseDbtProject):
...
+            if self.parent_project.name in project_dict.get(key, {}).keys():
+                project_dict[key][self.name] = project_dict[key].pop(self.parent_project.name)
+        project_dict["name"] = self.name
+        self.project = Project.from_dict(project_dict)
+
+    def _get_custom_macros(self) -> Set[str]:
+        """
+        get a set of macro unique_ids for all the selected resources
+        """
+        macro_list = []
```
nicholasyager, line 295: Since we're converting `macro_list` to a `Set` on line 308, it may make sense to initialize this as a set from the beginning. This will reduce the memory usage of `project_macros` since we're not storing duplicate values in `List`.

dbt_meshify/dbt_projects.py
nicholasyager: I can dig it!

dbt_meshify/dbt_projects.py
nicholasyager: Yeah, this feels like an odd level of abstraction. My mind goes to some other class that operates on `DbtProject`s to write files. Open to other approaches, though!

dbt_meshify/storage/yaml_editors.py
```diff
@@ -259,36 +335,81 @@ def add_model_version(
...
+
+    def move_resource(self):
+        """
+        move a resource file from one project to another
+
+        """
+        current_path = self.get_resource_path()
+        new_path = self.subdirectory / current_path
+        new_path.parent.mkdir(parents=True, exist_ok=True)
+        current_path.rename(new_path)
```
nicholasyager, line 376: My gut reaction is that this is a little odd. I suspect that it would be more ergonomic/useful to pass the destination path in as an argument. Having that been said, I'm only ~45% confident in this opinion.

dave-connors-3: i don't disagree! i think it's mostly dependent on how much flexibility we want to afford users on where resources land in the new project

dave-connors-3:
@nicholasyager agree on all fronts! made the small adjustments you recommended, and will spend some time today refactoring the initialization into a separate class that operates on dbt projects rather than have dbt projects have knowledge of the filesystem. 

re: logging -- couldn't agree more, I'll open up a new issue so we can tackle it separately for all the commands we have! would love a dbt-esque logging experience in here too

dave-connors-3:
@nicholasyager these tests pass locally -- I think my attempts to create and destroy test projects on the fly may need to be rethought!

in any case, this more or less is ready to be reviewed. I am ready to hand you the reigns on this PR to take a heavy editor's pen to the logic i've mangled here

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+    model_ordered_dict.update(model_yml)
+    # remove any keys with None values
+    return filter_empty_dict_items(model_ordered_dict)
+
+
+def resources_yml_to_dict(resources_yml: Dict, resource_type: NodeType = NodeType.Model):
+    """Converts a yml dict to a named dictionary for easier operation"""
+    return (
+        {resource["name"]: resource for resource in resources_yml[resource_type.pluralize()]}
+        if resources_yml
```
nicholasyager, line 43: Double check the type hints for `resource_yml`. If this variable can be false, then it's type cannot just be `Dict`. Maybe `Optional[Dict]`?

nicholasyager (commented):
Here are some light comments. I'm going to review the file editors more in depth soon:tm: once I have some more bandwidth.

dbt_meshify/dbt_projects.py
```diff
@@ -152,7 +159,19 @@ def get_catalog_entry(self, unique_id: str) -> Optional[CatalogTable]:
 
     def get_manifest_node(self, unique_id: str) -> Optional[ManifestNode]:
         """Returns the catalog entry for a model in the dbt project's catalog"""
-        return self.manifest.nodes.get(unique_id)
+        if unique_id.split(".")[0] in [
```
b-per, line 162: What about `exposure` and `metric`/`measure`? Are those not listed on purpose?

dbt_meshify/storage/dbt_project_creator.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def copy_packages_yml_file(self) -> None:
+        """
+        Writes the dbt_project.yml file for the subproject in the specified subdirectory
+        """
+        self.file_manager.copy_file(Path("packages.yml"))
+
+    def copy_packages_dir(self) -> None:
+        """
+        Writes the dbt_packages director to the subproject's subdirectory to avoid the need for an immediate deps run
```
b-per, line 70:
```suggestion
        Writes the dbt_packages directory to the subproject's subdirectory to avoid the need for an immediate `dbt deps` command
```

test-projects/split/split_proj/models/marts/__models.yml
```diff
@@ -2,7 +2,8 @@ version: 2
 
 models:
   - name: customers
-    description: Customer overview data mart, offering key details for each unique customer. One row per customer.
+    description: Customer overview data mart, offering key details for each unique
+      customer. One row per customer.
```
b-per, line 6: All the YAML files have some odd newlines showing up now.

tests/integration/test_split_command.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+
+        assert result.exit_code == 0
+        assert (
+            Path(dest_project_path) / "my_new_project" / "models" / "staging" / "stg_orders.sql"
+        ).exists()
+        x_proj_ref = "{{ ref('my_new_project', 'stg_orders') }}"
+        child_sql = (Path(dest_project_path) / "models" / "marts" / "orders.sql").read_text()
+        assert x_proj_ref in child_sql
+        teardown_test_project()
```
b-per, line 48: Would a teardown after an `assert` work as expected? I'd think that it would stop at the `assert` but it might just be my lack of `pytest` knowledge.

tests/unit/test_update_ref_functions.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return my_sql_model_df
+"""
+
+
+def read_yml(yml_str):
+    return yaml.safe_load(yml_str)
+
+
+class TestRemoveResourceYml:
```
b-per, line 39: We should have some tests when using model versions as well

tests/integration/test_subproject_creator.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def setup_new_project(write_packages_yml: bool = False):
+    dbt = Dbt()
+    subprocess.run(["poetry", "run", "dbt", "init", "test", "-s"])
+    with open("test/profiles.yml", "w") as f:
+        f.write(yaml.dump(test_project_profile))
+    if write_packages_yml:
+        with open("test/packages.yml", "w") as f:
+            f.write(yaml.dump(test_package_yml))
```
b-per, line 42: We should use `pathlib` like in the other parts of the code rather than `open`.

tests/integration/test_subproject_creator.py
```diff
@@ -0,0 +1,152 @@
...
+    if not resource:
+        raise KeyError(f"Resource {unique_id} not found in manifest")
+    meshify_constructor = DbtMeshConstructor(
+        project_path=subproject.path, node=resource, catalog=None
+    )
+    return meshify_constructor
+
+
+def teardown_new_project():
+    os.system("rm -rf test-projects/test")
```
b-per, line 63: 😉 let's use `shutil.rmtree` instead of `os.system`.

More generally, it is now recommended to use `subprocess.run` rather than `os.system` when running commands from Python

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+        else:
+            yml_path = Path(self.node.original_file_path)
+
+        # if the model doesn't have a patch path, create a new yml file in the models directory
+        if not yml_path:
+            resource_path = self.get_resource_path()
+
+            if resource_path is None:
+                # If this happens, then the model doesn't have a model file, either, which is cause for alarm.
+                raise Exception(f"Unable to locate the file defining {self.node.name}. Aborting")
```
nicholasyager, line 327: I'd love for this to raise a specific exception. Maybe a `ModelMissingError`, since we cannot recover from this type of issue.

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+        last_version_file_name = f"{self.node.name}_v{latest_version}.{self.node.language}"
+        next_version_file_name = (
+            f"{defined_in}.{self.node.language}"
+            if defined_in
+            else f"{self.node.name}_v{latest_version + 1}.{self.node.language}"
+        )
+        model_path = self.get_resource_path()
+
+        if model_path is None:
+            raise Exception(f"Unable to find path to model {self.node.name}. Aborting.")
```
nicholasyager, line 411: Should there ever be a time when `get_resource_path` returns None, or do we want this to error out? My gut says erroring out is the preferred approach, that way do don't introduce surface area for defects.

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+                model_code=model_code,
+            )
+        elif self.node.language == "python":
+            updated_code = self.update_python_refs(
+                model_name=model_name,
+                project_name=project_name,
+                model_code=model_code,
+            )
+        # write the updated model code to the file
+        self.file_manager.write_file(model_path, updated_code)
```
nicholasyager, line 453: Some simplification of our code paths:

```suggestion
        model_code = str(self.file_manager.read_file(model_path))
        # This can be defined in the init for this clas.
        ref_update_methods = {
            'sql': self.update_sql_refs,
            'python': self.update_python_refs
        }
        # Here, we're trusting the dbt-core code to check the languages for us. 🐉 
        updated_code = ref_update_methods[self.node.language](
                model_name=model_name,
                project_name=project_name,
                model_code=model_code,
        )
        # write the updated model code to the file
        self.file_manager.write_file(model_path, updated_code)
```

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+        remaining_tables = list(filter(lambda x: x["name"] != resource_name, tables))
+        resource_yml = source_definition.copy()
+        resource_yml["tables"] = table
+        source_definition["tables"] = remaining_tables
+        sources[source_name] = source_definition
+        if len(remaining_tables) == 0:
+            return resource_yml, None
+        else:
+            full_yml["sources"] = list(sources.values())
+            return resource_yml, full_yml
```
nicholasyager, line 131: The `else` is not required here.
```suggestion
        if len(remaining_tables) == 0:
            return resource_yml, None
 
        full_yml["sources"] = list(sources.values())
        return resource_yml, full_yml
```

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+    def get_source_yml_entry(
+        self, resource_name: str, full_yml: Dict[str, Any], source_name: str
+    ) -> Tuple[Dict[str, Any], Optional[Dict[str, Any]]]:
+        """
+        Remove a single source entry from a source defintion block, return source definition with single source entry and the remainder of the original
+        """
+        sources = resources_yml_to_dict(full_yml, NodeType.Source)
+        source_definition = sources.get(source_name)
+        tables = source_definition.get("tables", None)
+        table = list(filter(lambda x: x["name"] == resource_name, tables))
```
nicholasyager, line 121: Warning! A `None` type is not iterable.
```suggestion
        tables = source_definition.get("tables", [])
        table = list(filter(lambda x: x["name"] == resource_name, tables))
```

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,453 @@
...
+        """Add group and access configuration to a model's YAMl properties."""
+        # parse the yml file into a dictionary with model names as keys
+        models = resources_yml_to_dict(models_yml)
+        model_yml = models.get(model_name) or {"name": model_name, "columns": [], "config": {}}
+
+        model_yml.update({"group": group.name})
+        models[model_name] = process_model_yml(model_yml)
+
+        models_yml["models"] = list(models.values())
+        return models_yml
```
nicholasyager, line 100: This method, `add_access_to_model_yml`, and `add_group_to_yml` are _very_ similar in function -- specifically they're adding properties to a specific model property entry. This is fine for now, since this is a first draft. Having that been said, this might be something to add to the IO refactoring work I've been pondering.

nicholasyager (commented):
Added some more feedback here! 

My tl;dr of the file manager is that it is sufficient, but there are a few opportunities for simplification. I also found a couple areas where defects can trickle in (raising exceptions instead of returning None, using specific exceptions, etc). One major benefit of having this functionality written is that it makes clear to me where we need cleaner APIs -- specifically around manipulating resources. I wonder if a pre-v1 step will be to create first-class classes for our resources and manage SerDe. Definitely a task for later!

Omissions:
- I've not dug into the yml files generated yet.
- I'm going to trust the updated tests.

Open questions:
- How is mypy holding up? I suspect there are a few typing issues around None values and iterables.

dbt_meshify/dbt_projects.py
dave-connors-3: this logic was to get `nodes` vs the resources that are also other top-level keys in the manifest. Looking with fresh eyes, I would guess there's a dbt-core class that represents this type that we could leverage instead of a list of strings!

dbt_meshify/dbt_projects.py
```diff
@@ -152,7 +159,19 @@ def get_catalog_entry(self, unique_id: str) -> Optional[CatalogTable]:
 
     def get_manifest_node(self, unique_id: str) -> Optional[ManifestNode]:
         """Returns the catalog entry for a model in the dbt project's catalog"""
```
dave-connors-3, line 161:
```suggestion
        """Returns the manifest entry for a resource in the dbt project's manifest"""
```

tests/integration/test_split_command.py
dave-connors-3: it seems to work locally!

dbt_meshify/main.py
```diff
@@ -65,40 +68,36 @@ def connect(projects_dir):
...
 
 @cli.command(name="split")
+@create_path
+@click.argument("project_name")
 @exclude
 @project_path
 @select
 @selector
-def split():
+def split(project_name, select, exclude, project_path, selector, create_path):
```
nicholasyager, line 77: Since `select` now allows multiple arguments, we cannot have `--select` before the `project_name` argument.

```console
(dbt-meshify-py3.11) > $ poetry run dbt-meshify split --select "+orders" revenue                                                                                                                                         
Usage: dbt-meshify split [OPTIONS] PROJECT_NAME
Try 'dbt-meshify split --help' for help.

Error: Missing argument 'PROJECT_NAME'.
```

Instead, we need to order arguments/options specifically

```
(dbt-meshify-py3.11) > $ poetry run dbt-meshify split revenue --select "+orders"                                                                                                                                   
```

I don't this is a blocker per se. At the very least, documentation should be refined in a follow-up.

nicholasyager (approved):
What a chonky feature! I did another 🔍 of this PR and found a few small areas of refactoring and a CLI UX oddity. Having this been said, there were no flaming red flags along the way, so it's a ✅ from me! Let's get this out there so we can get feedback.

dbt_meshify/main.py
```diff
@@ -65,40 +68,36 @@ def connect(projects_dir):
...
-
-    print(holder.project_map())
+    subproject = project.split(
+        project_name=project_name, select=select, exclude=exclude, selector=selector
+    )
+    logger.info(f"Selected {len(subproject.resources)} resources: {subproject.resources}")
+    target_directory = Path(create_path) if create_path else None
+    subproject_creator = DbtSubprojectCreator(
+        subproject=subproject, target_directory=target_directory
+    )
```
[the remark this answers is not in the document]
dave-connors-3: that's good feedback, definitely agree -- IIRC the init handles this, but it's definitely not necessary to hide that behavior!

dbt_meshify/storage/file_content_editors.py
```diff
@@ -0,0 +1,466 @@
...
+            "columns",
+            "versions",
+        ]
+    )
+    model_ordered_dict.update(model_yml)
+    # remove any keys with None values
+    return filter_empty_dict_items(model_ordered_dict)
+
+
+def resources_yml_to_dict(resources_yml: Dict, resource_type: NodeType = NodeType.Model):
```
dave-connors-3, line 40:
```suggestion
def resources_yml_to_dict(resources_yml: Optional[Dict], resource_type: NodeType = NodeType.Model):
```

dbt_meshify/storage/file_content_editors.py
dave-connors-3: these were copy pasted in order to access each method separately, rather than only as part of `add_access_and_group_to_yml` -- I think the steps in all of these methods are overlapping and too verbose -- i want to shrink this file dramatically!

dbt_meshify/storage/file_content_editors.py
dave-connors-3: would that require a new class that extends `Exception`?

nicholasyager: Yep! Like a good candidate for a future refactor.

PR #73, opened 2023-06-26
add logging info

### LINKED ISSUE #68: add structured logs
### Describe the feature

Right now, you have no idea what's going on when you run a command! One should have an idea what is going on when one runs a command

do nothing

### Who will this benefit?

users!

maybe?

dave-connors-3:
resolves #68 

@nicholasyager tried my hand at using `loguru` for some basic logging messages here. very open to changing how we structure this and whether we use this package at all -- it's pretty straightforward to use, but I have never added logs to a cli app before, so I don't know what best practices to consider.

### DIFF
--- a/dbt_meshify/main.py
+++ b/dbt_meshify/main.py
@@ -136,7 +136,7 @@ def add_contract(select, exclude, project_path, selector, public_only=False):
             logger.success(f"Successfully added contract to model: {model_unique_id}")
         except Exception as e:
             logger.error(f"Error adding contract to model: {model_unique_id}")
-            logger.error(e)
+            logger.exception(e)
 
 
 @operation.command(name="add-version")
@@ -172,7 +172,7 @@ def add_version(select, exclude, project_path, selector, prerelease, defined_in)
                 logger.success(f"Successfully added version to model: {model_unique_id}")
             except Exception as e:
                 logger.error(f"Error adding version to model: {model_unique_id}")
-                logger.error(e)
+                logger.exception(e)
 
 
 @operation.command(name="create-group")
@@ -234,7 +234,7 @@ def create_group(
         logger.success(f"Successfully created group: {name}")
     except Exception as e:
         logger.error(f"Error creating group: {name}")
-        logger.error(e)
+        logger.exception(e)
 
 
 @cli.command(name="group")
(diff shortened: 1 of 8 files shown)


dbt_meshify/main.py
```diff
@@ -105,22 +111,32 @@ def add_contract(select, exclude, project_path, selector, public_only=False):
...
         model_catalog = project.get_catalog_entry(model_unique_id)
         meshify_constructor = DbtMeshModelConstructor(
             project_path=project_path, model_node=model_node, model_catalog=model_catalog
         )
-        meshify_constructor.add_model_contract()
+        logger.info(f"Adding contract to model: {model_unique_id}")
+        try:
+            meshify_constructor.add_model_contract()
+            logger.success(f"Successfully added contract to model: {model_unique_id}")
+        except Exception as e:
```
b-per, line 137: I am OK with this for now to add debugging but I have 2 comments:
- what do we want the tool to do when there is an error? stop at the first error or skip the error and continue until the end (which is what we are doing with the `try/except`)
- if possible, we should be specific about the exceptions raised so we can add specific logic/logging: `Exception as e` is the generic one but we could try to be more specific. This is something that can be revisited later.

b-per (approved):
OK as is but I added 1 comment.

dbt_meshify/storage/yaml_editors.py
```diff
@@ -248,14 +251,18 @@ def add_model_version(
...
+                model_name=self.model_node.name,
+                models_yml=models_yml,
+                prerelease=prerelease,
+                defined_in=defined_in,
+            )
+            # write the updated yml to the file
+            self.file_manager.write_file(yml_path, updated_yml)
+            logger.info("Model version added to model yml")
+        except Exception as e:
+            logger.error(f"Error adding model version to model yml: {e}")
```
nicholasyager, line 265: Currently, this try/except eats the exception. Am I correct in believe that we want to re-raise the exception after logging?

nicholasyager (commented):
Two main areas for feedback/discussion include:
1. CLI UX for setting debug/log-level.
2. Exceptions being swalloed by the try/except behavior.

dave-connors-3:
@nicholasyager @b-per made a couple updates yesterday that I think address the question of try/except -- in the except branch, i now include `logger.exception(e)`, which I still raises the exception, as Nich mentioned in their comment. Let me know if that is not sufficient! 

@b-per -- re: more detailed excpetions, i totally agree, but wasn't positive which class was more representative of what the error might be at the level I placed the logs.

dave-connors-3:
@nicholasyager @b-per i am going to merge this as is so I can add logging to PR #63 before I leave to go OOO. I opened #92 to reflect the outstanding discussion points from above!

PR #81, opened 2023-06-26
raise exception if no selection syntax provided to group command

### LINKED ISSUE #77: Default create-group with nothing selected, makes a group with all of your models
### Describe the bug

When I run `dbt-meshify operation create-group raw_vault`, dbt-meshify creates a new group with _all_ of my project's models in it. I would expect that when I don't specify selector syntax, dbt-meshify would create a new _empty_ group.

Thoughts?

dave-connors-3:
closes #77 

if no argument provided for selecting nodes, the group command will fail!

### DIFF
--- a/dbt_meshify/main.py
+++ b/dbt_meshify/main.py
@@ -178,6 +178,11 @@ def create_group(
     """
     from dbt_meshify.utilities.grouper import ResourceGrouper
 
+    if not (select or exclude or selector):
+        raise Exception(
+            "Groups must be a subset of the resources in your project. Please provide a select, exclude, or selector argument to filter the resources in the project."
+        )
+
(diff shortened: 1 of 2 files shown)


graciegoheen:
Why have it fail vs. create an empty group?

dave-connors-3:
open to it! what's the benefit of an empty group?

graciegoheen:
Maybe I want to start by creating an empty group and manually add models to it

dave-connors-3:
@nicholasyager @graciegoheen updated the logic to simply create an empty group without selection syntax!

nicholasyager (approved):
@dave-connors-3 This change looks both necessary and sufficient. Kudos for catching the test updates, too!

PR #83, opened 2023-06-27
add read catalog step if catalog exists

### LINKED ISSUE #80: read `catalog.json` if it already exists rather than a new `dbt docs generate` on each command
### Describe the feature

Right now, each invocation loads a catalog using a docs generate command. For large projects, this stinks and is very slow. we should load the `catalog.json` file from the target directory if it exists, and only generate a new one if it's not already there.

Leaving it alone! 

### Who will this benefit?

Users with large dbt projects

Yes

dave-connors-3:
closes #80 

This PR adds a step to the `DbtProject.from_directory()` method to first check if there's a `catalog.json` in the target directory, and uses it if so. If not, it will call the docs generate command using the dbt runner. 

I would love advice on how to test this -- i don't think we have a test suite for the `DbtProject` class at the moment!

## Update 7/5

Added a CLI flag (`--read-catalog`, or `-r` , or using envvar `DBT_MESHIFY_READ_CATALOG` ) to optionally read an existing `catalog.json` file. The default behavior is still to run a docs generate. If the flag is used, and no catalog exists, we fall back to running a docs generate.


b-per:
This would be great for projects with a slow-ish `dbt docs generate`!

I know it's in Draft but a couple of comments:

- we'd need to have a CLI flag to be able to deactivate reading from JSON. Some people might have an "old" `catalog.json` that has not been updated and that people would like not to take into account
- it currently expects the `catalog.json` to be in `target`, that's fair, but what about people who might download it from dbt Cloud? Would we always want to have it in `target` or allow any path?

moreaupascal56:
Hello, first thanks for the package I just discovered it and it seems very useful! 

Just a remark, don't you think the `dbt docs generate` option should be the default, and the CLI flag should be able to activate the reading of a JSON existing file (instead of deactivate)? 

So when running this command with no flags at all you are sure that the data is up to date. 
Just a thought because my catalog is always outdated but I may not be an example ! 
Have a great day !

dbt_meshify/dbt_projects.py
```diff
@@ -190,15 +188,29 @@ def _load_project(path) -> Project:
...
+            if read_catalog:
+                logger.info(f"Reading catalog from {catalog_path}")
+                try:
+                    catalog_dict = json.loads(catalog_path.read_text())
+                    return CatalogArtifact.from_dict(catalog_dict)
+                except FileNotFoundError:
+                    logger.info(f"Catalog not found at {catalog_path}, running dbt docs generate")
+                    return dbt.docs_generate(directory)
+            else:
+                return dbt.docs_generate(directory)
```
nicholasyager, line 208: Non-blocking idea: Make this method a static function a class method or a static function.

graciegoheen (approved):

PR #91, opened 2023-07-05
fix version bug when incrementing versions on a prereleased model

### LINKED ISSUE #90: duplicate versions defined after using `--prerelease` flag for version operation
### Describe the bug

when using the `add-version` operation, if you use the `--prerelease` flag, which skips the `latest_version:` increment, and creates a file, the latest version and the defined versions are intentionally our of sync. Subsequent invocations of the add-version command after having used the --pre-release flag result in duplicated version numbers in yml. 

### Steps to reproduce

1. invoke `dbt-meshify operation add-version -s my_model`

yml result:

```
models: 
```
  - name: my_model
    latest_version: 1
    versions:
      - v: 1
```

2. invoke `dbt-meshify operation add-version -s my_model --prerelease`

yml result:

```
models: 
```
  - name: my_model
    latest_version: 1
    versions:
      - v: 1
      - v: 2
```
no error

1. invoke `dbt-meshify operation add-version -s my_model`

yml result:

```
models: 
  - name: my_model
    latest_version: 2
    versions:
      - v: 1
      - v: 2
      - v: 2
```
duplicate version entry

### Expected results

The add version command should
1. (no prerelease flag) increment `latest_version` and create next version + file based on defined yml versions
2. (prerelease flag) _not_ increment `latest_version` and still create next version + file based on defined yml versions

no parse errors from dbt

### Actual results

dupe versions and a parse error to boot

### System information
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

```
<output goes here>
```

```
<output goes here>
```

dave-connors-3:
Closes #90 

Rather than relying on `latest_version` from the model yml to determine the filepath of the next version, we should use the defined versions in yml, since latest_version may not equal the last version for a model defined.


rijnhardtkotze:
Hey @dave-connors-3

AFAIK from the schema of the ModelNode, the `version` can be a string as well. We are assuming that it is a number in this PR.

Will that always be the case?

dave-connors-3:
@rijnhardtkotze fantastic question! i was under the impression that these will always be integer values, but you're 100% correct that string values are allowed here -- let me ask around internally about why we're allowing strings. If there's good reason, I can adjust this behavior to handle those cases!

rijnhardtkotze:
It also seems weird to me that strings are allowed in versions. 

I can imagine it being quite weird when you define a version `foo` on model `bar`. 

The end result would look out of place with the floating `v` in there. `bar_vfoo`.

Perhaps this is a discussion for dbt-core?

graciegoheen:
@rijnhardtkotze I believe that core is not strict about requiring numeric versions, because users may be interested in using non-integer versions for their models (such as semantic versioning e.g. 1.0.0).

That being said, I think it makes sense for this package to use integer-only versions - in an attempt to enforce "best practices". 

I'm going to go ahead an approve & merge this PR, but if this becomes an issue in the future we can re-visit :)

Issue #93, opened 2023-07-07
update docs to reflect multiselect behavior of select


dave-connors-3:
Since `select` now allows multiple arguments, we cannot have `--select` before the `project_name` argument.

```console
(dbt-meshify-py3.11) > $ poetry run dbt-meshify split --select "+orders" revenue                                                                                                                                         
Usage: dbt-meshify split [OPTIONS] PROJECT_NAME
Try 'dbt-meshify split --help' for help.

Error: Missing argument 'PROJECT_NAME'.
```

Instead, we need to order arguments/options specifically

```
(dbt-meshify-py3.11) > $ poetry run dbt-meshify split revenue --select "+orders"                                                                                                                                   
```

I don't this is a blocker per se. At the very least, documentation should be refined in a follow-up.

_Originally posted by @nicholasyager in https://github.com/dbt-labs/dbt-meshify/pull/63#discussion_r1255963088_

PR #97, opened 2023-07-11
fix(cli): Create custom Command type to support custom Usage docs

nicholasyager:
Resolves: #93

# Description

This PR implements a custom Click Command type called `TupleCompatibleCommand`, which has a custom usage string implementation which places the `[OPTION]` metavar at the end of the usage string. 

**Original**
```console
dbt-meshify split [OPTIONS] PROJECT_NAME
```

**Updated**
```console
dbt-meshify split PROJECT_NAME [OPTIONS]
```

This is important from a UX perspective, since `--select` and other type-typed options will eat all subsequent strings that are not themselves options. For example, 

```console
dbt-meshify split --select foo bar
```
will yield an error, since `bar` will be interpreted as part of the selector instead of as the project name.

Now, if users follow the example, we will have the following correct usage:

```console
dbt-meshify split bar --select foo
```

### DIFF
--- a/dbt_meshify/cli.py
+++ b/dbt_meshify/cli.py
@@ -1,6 +1,7 @@
 import functools
 
 import click
+from click import Context, HelpFormatter
 from dbt.cli.options import MultiOption
 
 # define common parameters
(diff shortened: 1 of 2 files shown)


dave-connors-3 (approved):
🦞

Issue #94, opened 2023-07-07
create specific errors for missing files


dave-connors-3:
I'd love for this to raise a specific exception. Maybe a `ModelMissingError`, since we cannot recover from this type of issue.

_Originally posted by @nicholasyager in https://github.com/dbt-labs/dbt-meshify/pull/63#discussion_r1251027938_

PR #100, opened 2023-07-18
Feature: Add new exceptions for file management

nicholasyager:
Resolves: #94 

# Description and motivation
This PR implements two new Exceptions to the code base for more descriptive error handling when required. These are:
1. `FileEditorException`: Exceptions raised by a file editor during self-checks. Currently, this applies when a YAML reader returns a `str` instead of a `Dict`.
2. `ModelFileNotFoundError`: A fatal error that occurs when a model is missing both a patch path and an original file path. This is particularly tricky, since it means that we _do not_ have an actual source for the model.

Along the way, I created an `exceptions.py` file to store these exceptions, and I moved the pre-existing `FatalMeshifyException` into this file.

### DIFF
--- a/dbt_meshify/exceptions.py
+++ b/dbt_meshify/exceptions.py
@@ -1,0 +1,22 @@
+import click
+from loguru import logger
+
+
+class FileEditorException(BaseException):
+    """Exceptions relating to errors in file generation and loading."""
+
+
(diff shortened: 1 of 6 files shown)


dave-connors-3 (approved):
LGTM! 🦞

PR #103, opened 2023-07-24
Fix create path

### LINKED ISSUE #101: `--create-path` arg doesn't actually accept a path
### Describe the bug

the `--create-path` option for the split command (which is not well named) is intended to be the name of the subdirectory created in the case when the user wants a folder name other than the name of the project itself. This arg is also likely to be used when a user wants to create a new dbt project parallel to the parent project rather than nested within. This argument should be refactored to support real paths. 

### Steps to reproduce

```
dbt-meshify split bread --create-path ~/user/dev/baked_goods -s tag:loaf
```

will run into a fatal exception. 

### Expected results

It Just Works

### Are you interested in contributing the fix?
sure

dave-connors-3:
Closes #101 

This PR makes the `create-path` CLI arg supplied to the `split` command, well, work! It now accepts any direct or relative path, allowing end users to split their projects into any directory they'd like!

### DIFF
--- a/dbt_meshify/cli.py
+++ b/dbt_meshify/cli.py
@@ -14,7 +14,7 @@ project_path = click.option(
 
 create_path = click.option(
     "--create-path",
-    type=click.Path(exists=True),
+    type=click.Path(exists=False),
     default=None,
     help="The path to create the new dbt project. Defaults to the name argument supplied.",
 )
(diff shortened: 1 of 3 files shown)


nicholasyager (approved):
@dave-connors-3 This change is both necessary and sufficient. Kudos for adding tests. Great work ✅

PR #106, opened 2023-07-28
Fix overwriting default yml path

### LINKED ISSUE #105: un-yml'd models overwrite default yml path
### Describe the bug

Right now, if a model does not a have a yml entry, we default it to `path/to/model/_models.yml`. Without checking, we create this file. That means if that file already exists with other model entryies, and we run an operation for a model in that directory, we will replace the file with the resulting yml from the operation rather than adding to the existing yml. 

Should be an easy oneliner!

### Steps to reproduce

1. open sample project with no yml and two models in the same subdirectory 
2. run `dbt-meshify operation add-contract -s model_1`
3. observe new yml in new file next to model_1
4. run `dbt-meshify operation add-contract -s model_2`
5. obeserve the yml from step 2 is overwritten with the yml from step 4

### Expected results

additive operations, not destructive!

### Actual results

destructive operations, not additive!

### System information
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

```
<output goes here>
```

```
<output goes here>
```

dave-connors-3:
closes #105 

Checks that the yml file for a non-yml'd model exists before overwriting it!


nicholasyager (approved):
As discussed in our meeting, this PR looks ready for prime-time ✅

dave-connors-3:
closed in favor of #109

PR #110, opened 2023-08-08
update logic to accomodate child-only subprojects

dave-connors-3:
This turned into a bigger one than anticipated!

## Problem
The current `split` command checks for children of the selected resources, and updates those refs to select from the newly generated project. What it does _not_ do is update the _parents_ of the selected nodes, and apply the same checks (contract, make public those models, and update the refs within the selected models to refer to the "parent" project)

## Solution
When a subproject is created, list out the nodes that are cross project parents and children of the selected resources:

1. If there are both cross project parents and children of the selected nodes **_stop execution_** -- per our best practices, we should not allow for project cycles like this:
image

`a --> b --> a` is no good!

2. proceed as normal for boundary nodes - cross project children of selected resources get their sql updated to refer to the newly created project
3. for any selected resource that depends on one of the cross project parents, update its SQL to refer to the models in the parent project
4. for the non-selected resources that are the parents of the selected resources, add contracts and access to allow the new project to select from the across projects.

### DIFF
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -292,12 +292,55 @@ class DbtSubProject(BaseDbtProject):
         self.manifest = copy.deepcopy(parent_project.manifest)
         self.project = copy.deepcopy(parent_project.project)
         self.catalog = parent_project.catalog
-        self.custom_macros = self._get_custom_macros()
-        self.groups = self._get_indirect_groups()
-
-        self._rename_project()
 
         super().__init__(self.manifest, self.project, self.catalog, self.name)
+
+        self.custom_macros = self._get_custom_macros()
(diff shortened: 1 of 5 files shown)


nicholasyager (approved):
@dave-connors-3 thanks for tackling this so quickly! I've confirmed that this correctly resolves the leaf-splitting defect we were observing, so I'm a happy camper.

I have a few recommendations re: mypy issues and which dbt project-adjacent class should handle cross-project references. Having that been said... I'm happy to tackle some of these changes as part of merge conflict resolution on the `ChangeSet` PR.

dave-connors-3:
@nicholasyager If you're game, I may merge as is, and let the `ChangeSet` work tackle the clean up as fast follows!

nicholasyager:
That sounds entirely reasonable to me! 👍

PR #123, opened 2023-08-18
Missing yml entry error

### LINKED ISSUE #122: non-yml'd models throw `FileNotFoundError` in `DbtSubprojectCreator.move_resource_yml_entry`
### Describe the bug

in an effort to test the issue described in #121, i added a leaf node to the test project. I did not add a yml entry, and got a `FileNotFoundError`

### Steps to reproduce

1. add a model without a yml entry to the project
2. select it in the `split` command selector
3. see `Error creating subproject`

```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/daveconnors/dev/dbt-labs/dbt-meshify/test-projects/split/split_proj/models/_models.yml'
```

### Expected results
Existence of yml entry does not affect execution of the command

### Actual results
An error!

### System information
- [x] 🦆 

```
<output goes here>
```

```
<output goes here>
```

dave-connors-3:
Closes #122 

- only tries to load yml from files that exist
- only adds changes for moving yml entries that exist

### DIFF
--- a/dbt_meshify/storage/dbt_project_editors.py
+++ b/dbt_meshify/storage/dbt_project_editors.py
@@ -277,27 +277,26 @@ class DbtSubprojectCreator:
         resource_entry = self.load_resource_yml(
             current_yml_path, resource.name, resource.resource_type, source_name
         )
-        if not resource_entry:
-            return change_set
-        change_set.extend(
-            [
-                ResourceChange(
-                    operation=Operation.Add,
-                    entity_type=EntityType(resource.resource_type.value),
-                    identifier=resource.name,
-                    path=new_yml_path,
-                    data=resource_entry,
-                    source_name=source_name,
-                ),
-                ResourceChange(
-                    operation=Operation.Remove,
-                    entity_type=EntityType(resource.resource_type.value),
-                    identifier=resource.name,
-                    path=current_yml_path,
-                    data={},
-                    source_name=source_name,
-                ),
-            ]
-        )
+        if resource_entry:
+            change_set.extend(
+                [
+                    ResourceChange(
+                        operation=Operation.Add,
+                        entity_type=EntityType(resource.resource_type.value),
+                        identifier=resource.name,
+                        path=new_yml_path,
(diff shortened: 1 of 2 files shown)


dbt_meshify/storage/dbt_project_editors.py
```diff
@@ -279,7 +277,8 @@ def move_resource_yml_entry(self, resource: Resource) -> ChangeSet:
         resource_entry = self.load_resource_yml(
             current_yml_path, resource.name, resource.resource_type, source_name
         )
-
+        if not resource_entry:
```
dave-connors-3, line 280: you know i wrote this thinking about the earlier feedback about returning early but i almost feel like only extending the change set if there is a resource entry might actually read nicer here

nicholasyager: >i almost feel like only extending the change set if there is a resource entry might actually read nicer here

I'm inclined to agree 👍🏻

dbt_meshify/storage/dbt_project_editors.py
```diff
@@ -48,7 +48,7 @@ def load_resource_yml(
         self, path: Path, name: str, resource_type: NodeType, nested_name: Optional[str] = None
     ) -> Dict:
         """Load the Model patch YAML for a given ModelNode."""
-        raw_yml = YAMLFileManager.read_file(path)
+        raw_yml = YAMLFileManager.read_file(path) if path.exists() else {}
```
nicholasyager, line 51: Great change! I suspect an alternative approach would be to handle the `FileNotFoundError`, but this seems a bit more elegant since we know to check for the path first.

nicholasyager (approved):
This PR looks great! I agree that it may be better to not return early in this specific case, but what you have written is both necessary and sufficient. ✅

dbt_meshify/storage/dbt_project_editors.py
dave-connors-3: updated!

PR #130, opened 2023-08-18
cast all relation names to lowercase while evaluating dependencies

### LINKED ISSUE #126: connect command should not be case sensitive
### Describe the bug

I tried to run the `dbt-meshify connect` command on two projects connected via the "source hack". 

I got this message:
Screenshot 2023-08-18 at 2 20 19 PM

After investigating, it was because my target schema in my upstream project was ALL CAPS which my source definition in my downstream project was lower case.

dave-connors-3:
Closes #126 

Pretty straighforward if you ask me!


nicholasyager (approved):
A no-brainer change. Thanks @dave-connors-3 

Make it so

Issue #129, opened 2023-08-18
source not fully deleted if no longer needed


dave-connors-3:
### Describe the bug

when using the `connect` command, if the source entry is removed, and is the only source entry, the entire file should be removed. Currently, it writes an empty array to the `tables` key of the yml dictionary

<img width="773" alt="image" src="https://github.com/dbt-labs/dbt-meshify/assets/73915542/eb3353f7-958a-4933-8a0f-e6038c584062">

### Steps to reproduce

in our `test-projects/source-hack` directory, run

```bash
dbt-meshify connect --project-paths src_proj_a src_projb
```

### Expected results

File is deleted if no sources remain

### Actual results

see above

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt debug`:**
```
<output goes here>
```

**The output of `dbt --version`:**
```
<output goes here>
```

PR #131, opened 2023-08-20
Fix: YAMLFileManager removes dictionary fields containing empty lists

nicholasyager:
# Description and motivations
This PR adds some extra content awareness to `YMLFileManager` so that it will check for Dict fields with empty lists, and remove those fields prior to serialization. This is important to prevent unnecessary fields from lingering in YAML files after dbt-meshify modifies file content. 

Resolves: #129


dave-connors-3:
This looks good to me -- just to be clear on the behavior of this fix, would removing `table_1` from this source:

```yml

sources:
  - name: upstream_modeal
    database: analytics
    schema: prod
    tables:
       - name: table_1

```

result in this yml?

```yml

sources:
  - name: upstream_modeal
    database: analytics
    schema: prod

```

Thinking specifically about the `dbt-meshify connect` case, I wonder if the behavior we'd actually want is to delete the entire parent source entry altogether? Would feel a little awkward to leave half a source definition in place!

nicholasyager:
> This looks good to me -- just to be clear on the behavior of this fix, would removing `table_1` from this source:
> 
> ```yaml
> sources:
>   - name: upstream_modeal
>     database: analytics
>     schema: prod
>     tables:
>        - name: table_1
> ```
> 
> result in this yml?
> 
> ```yaml
> sources:
>   - name: upstream_modeal
>     database: analytics
>     schema: prod
> ```
> 
> Thinking specifically about the `dbt-meshify connect` case, I wonder if the behavior we'd actually want is to delete the entire parent source entry altogether? Would feel a little awkward to leave half a source definition in place!

That's a great idea. I can update the ResourceEditor to handle this case, too!

dave-connors-3 (approved):
rock and also roll

PR #144, opened 2023-08-21
clean subgraph before identifying boundary in split operation

dave-connors-3:
This PR adds a line to clean the subproject's graph of test nodes before identifying the boundary nodes. 

This operation is called during the creation of a group, but not splitting a project, resulting in divergent behavior. 

to test this, I added a tested leaf node to the project, and added a test case that splits that node out and makes sure it's public. I did a lazy version of checking the yml -- @nicholasyager should that be a named list?


nicholasyager (approved):
@dave-connors-3 This approach is both necessary and sufficient for removing tests from the boundary detection used in splitting. The test is a _little_ lazy, but it's also effective, so I'm cool with it 👍🏻 ✅

PR #145, opened 2023-08-22
build and leverage child and parent maps to properly evaluate cross project dependencies

### LINKED ISSUE #143: `update_child_refs` failure - evaluating each cross project children on each iteration
### Describe the bug

`ReferenceUpdater.update_child_refs` selects too many changes and can overwrite changes. 

Hypothesis is that the regex logic fails when models are the children pf multiple upstream boundary nodes, and there are conflicting changes in the change sets. 

These are unsubstantiated, but i think we have two issues

1. updating all downstream children all the time
[this line](https://github.com/dbt-labs/dbt-meshify/blob/ignore-tests-from-leaf-boundary-node-calc/dbt_meshify/utilities/references.py#L136) indicates we iterate across all children of the subproject _every time we call this method, regardless of the resource we pass in_. That's plainly wrong, and I believe to be why we saw no change to the code -- when a cross project child is called in this method when it does not actually depend on the `resource` we'll generate a project change with the `code` containing the post-processed code that didn't have any matching refs and therefore no updates

2. chilcren that have multiple boundary parents -- My hunch is that we would only end up with the _last_ parent's changes if we called multiple code changes on one downstream node for each individual parent. That seems harder and may make a new issue!

dave-connors-3:
resolves #143 

I will also split an additional issue out to track the question of "what happens if a downstream node has two cross-project parents?

this PR adds:
- logic to build the `child_map` and `parent_map` objects into dbt projects
- uses those maps to simplify and clarify the logic that detects cross project parents and children
- updates the logic in `update_child_refs` to _only evaluate children of the model called in the split operation_ (the original motivation of this PR!)

### DIFF
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -60,6 +60,9 @@ class BaseDbtProject:
         }
 
         self._graph = None
+        self.child_map: Dict[str, Set[str]] = {}
+        self.parent_map: Dict[str, Set[str]] = {}
+        self._build_parent_and_child_maps()
 
         self._changes: Dict[str, str] = {}
 
@@ -72,22 +75,27 @@ class BaseDbtProject:
         )
 
     def _get_xproj_children_of_selected_nodes(self) -> Set[str]:
-        return {
-            model.unique_id
-            for model in self.models.values()
-            if any(parent for parent in model.depends_on.nodes if parent in self.resources)
-            and model.unique_id not in self.resources
-        }
+        all_children: Set[str] = set()
+        for resource in self.resources:
+            if resource.startswith("test"):
+                continue
+            all_children.update(self.child_map.get(resource, {}))
+        return all_children - self.resources
 
     def _get_xproj_parents_of_selected_nodes(self) -> Set[str]:
-        resources = [self.get_manifest_node(resource) for resource in self.resources]
-        return {
-            node
-            for resource in resources
-            if resource is not None and isinstance(resource, (ModelNode, SnapshotNode))
-            for node in resource.depends_on.nodes
-            if node not in self.resources
-        }
+        all_parents: Set[str] = set()
+        for resource in self.resources:
+            if resource.startswith("test"):
+                continue
+            all_parents.update(self.parent_map.get(resource, {}))
+        return all_parents - self.resources
+
+    def _build_parent_and_child_maps(self) -> None:
+        if hasattr(self.manifest, "child_map") and hasattr(self, "parent_map"):
+            return
+        self.manifest.build_parent_and_child_maps()
+        self.child_map = self.manifest.child_map
+        self.parent_map = self.manifest.parent_map
 
     def _check_is_parent_of_parent_project(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


dbt_meshify/utilities/references.py
```diff
@@ -133,7 +133,10 @@ def update_child_refs(self, resource: CompiledNode) -> ChangeSet:
 
         change_set = ChangeSet()
 
-        for model in self.project.xproj_children_of_resources:
+        # import ipdb; ipdb.set_trace()
```
dave-connors-3, line 136:
```suggestion

```

dbt_meshify/utilities/references.py
```diff
@@ -133,7 +133,9 @@ def update_child_refs(self, resource: CompiledNode) -> ChangeSet:
 
         change_set = ChangeSet()
 
-        for model in self.project.xproj_children_of_resources:
+        for model in self.project.child_map[resource.unique_id]:
+            if model in self.project.resources:
+                continue
```
nicholasyager, line 138: This is such a better solution!

dbt_meshify/dbt_projects.py
```diff
@@ -72,22 +75,27 @@ def __init__(
...
+        all_parents: Set[str] = set()
+        for resource in self.resources:
+            if resource.startswith("test"):
+                continue
+            all_parents.update(self.parent_map.get(resource, {}))
+        return all_parents - self.resources
+
+    def _build_parent_and_child_maps(self) -> None:
+        if hasattr(self.manifest, "child_map") and hasattr(self, "parent_map"):
+            return
```
nicholasyager, line 95: I'm not sure that I understand what this logic is here for. Can I get a quick explanation?

dbt_meshify/dbt_projects.py
```diff
@@ -72,22 +75,27 @@ def __init__(
...
-            model.unique_id
-            for model in self.models.values()
-            if any(parent for parent in model.depends_on.nodes if parent in self.resources)
-            and model.unique_id not in self.resources
-        }
+        all_children: Set[str] = set()
+        for resource in self.resources:
+            if resource.startswith("test"):
+                continue
+            all_children.update(self.child_map.get(resource, {}))
```
nicholasyager, line 82: I'm not sure if `{}` evaluates as an empty set, or an empty dict. Might want to confirm this.

nicholasyager (commented):
@dave-connors-3 I feel like this is a solid approach, but I have one question about return logic on `build_parent_and_child_maps`, and another low-level question about `{}`.

dbt_meshify/dbt_projects.py
dave-connors-3: this logic looks a little drunk tbh -- let me double check whether this was a type check thing that may be eschewed altogether

dave-connors-3: I removed the `self.child_map = {}` and `self.parent_map = {}` from the `init` method, and removed this. IIRC i was battling with mypy when i was writing this manifest method in a slightly different order. I think this should be sufficient!

dbt_meshify/dbt_projects.py
dave-connors-3: in fact, the maps return lists! updated the types accordingly, and confirmed that a list is an acceptable arg to the `Set.update()` method

nicholasyager (approved):
@dave-connors-3 Thanks for simplifying this 👍🏻 ✅

Issue #119, opened 2023-08-17
`dbt-meshify operation create-group` sets model's access to public, should it instead be protected?


graciegoheen:
### Describe the bug

When I run `dbt-meshify operation create-group` to create a new group...
- all models that are only referenced by other models in their _same group_ will have `access: private`
- all other models (those that are referenced by models _outside their group_ or are leaf nodes) will have `access: public`

Is this correct? We're assuming, then, that this person has plans to split apart their project. But what if that's not the case? If you're just trying to create a group, wouldn't you want...
- all models that are only referenced by other models in their _same group_ will have `access: private`
- all models that are leaf nodes will have `access: public`
- all other models (those that are referenced by models _outside their group_) will have `access: protected`

Same question for the `dbt-meshify group` command.

### Steps to reproduce

- run the `dbt-meshify operation create-group` command on a subset of your project that has leaf nodes and nodes that are referenced by models outside the group

### Expected results

- models in the new group that are referenced by other models outside the group, would have access `protected`

### Actual results

- models in the new group that are referenced by other models outside the group, have access `public`

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake
- [ ] other (specify: ____________)

**The output of `dbt debug`:**
```
<output goes here>
```

**The output of `dbt --version`:**
```
<output goes here>
```

graciegoheen:
I don't know in what command this belongs - but what about updating the access of models upstream of exposures to be public and contracted?

PR #150, opened 2023-08-25
Fix: Update default access used in group creation to `protected`

nicholasyager:
# Description and motivation
In issue #119, we discussed that it doesn't make sense for the default access level granted during grouping to be `public`, since that would provide unexpected cross-project access. This PR updates the default access level granted to instead be `protected`. This will ensure that all models in the project can access the model, without granting cross-project access.

Resolves: #119

### DIFF
--- a/dbt_meshify/main.py
+++ b/dbt_meshify/main.py
@@ -464,7 +464,7 @@ def group(
                 for change in group_changes[0]
                 if isinstance(change, ResourceChange)
                 and change.entity_type == EntityType.Model
-                and change.data["access"] == "public"
+                and change.data["access"] != "private"
             ]
         ),
         project_path=project_path,
--- a/dbt_meshify/main.py
+++ b/dbt_meshify/main.py
@@ -444,17 +444,17 @@ def group(
 ) -> List[ChangeSet]:
     """
     Creates a new dbt group based on the selection syntax
-    Detects the edges of the group, makes their access public, and adds contracts to them
+    Detects the edges of the group, makes their access protected, and adds contracts to them
     """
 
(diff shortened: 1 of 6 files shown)


dbt_meshify/main.py
```diff
@@ -444,17 +444,17 @@ def group(
...
-    # that have a public access configuration on the boundary of our group. But, we don't
+    # that have a non-private access configuration on the boundary of our group. But, we don't
     # actually want to update our manifest yet. This puts us between a rock and a hard place.
     # to work around this, we can "trust" that create_group will always return a change for
-    # each public model, and rewrite our selection criteria to only select the models that
-    # will be having a public access config.
+    # each accessible model, and rewrite our selection criteria to only select the models that
+    # will be having a non-private access config.
 
     contract_changes = ctx.invoke(  # noqa: F841
```
dave-connors-3, line 459: This brings up an interesting point! in my mind, contracts and public models are pretty synonymous, but i hadn't thought about a contract on a protected model. I suppose in the context of a single project, if these are indeed the interfaces between groups, this makes sense, since the effect of grouping a monolith should probably be the same as splitting a project, but I can anticipate some questions about this choice!

dave-connors-3 (approved):
one open ended comment about contracts, but this looks good to me!

dbt_meshify/main.py
nicholasyager: Yeah, there will def be some questions!

IMHO, it all comes down to what the interfaces are. As you've stated, in a one-project (multi-hundred model and multi-AE team) world, the interfaces between groups ought to be contracted so that teams/groups can operate independently. But! We don't necessarily want to default to public access from other projects => protected models should be contracted when they're an interface.

Issue #138, opened 2023-08-21
`require-dbt-version` improperly handled on a split


dave-connors-3:
### Describe the bug

splitting a project with `require-dbt-version` doesn't properly parse the semantic version into the new project

<img width="300" alt="image" src="https://github.com/dbt-labs/dbt-meshify/assets/73915542/11e9c9a7-bf39-462a-ac29-819df5a01c06">

### Steps to reproduce

1. define a `require-dbt-version`
2. perform a split
3. observe `require-dbt-version` in newly created project

### Expected results

settings in both projects are equivalent

### Actual results

broke

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [x] snowflake
- [ ] other (specify: ____________)

**The output of `dbt debug`:**
```
<output goes here>
```

**The output of `dbt --version`:**
```
<output goes here>
```

nicholasyager:
l
o
o
k
s

f
i
n
e

t
o

m
e
!

PR #151, opened 2023-08-25
Fix: `require-dbt-versions` no longer serialized as a list of characters

nicholasyager:
# Description and motivations
The internal `Project` representation of the `require-dbt-versions` field within `dbt-core` is a list of strings, where each part of the version is an element in the list. When serialized in `write_project_file`, the resulting YAML is _also_ a list of strings. This PR adds some formatting to the `write_project_file` function to join the elements of the list into one complete string.

Resolves: #138 

# Example

## Input
```yml
# Name your project! Project names should contain only lowercase characters
# and underscores. A good package name should reflect your organization's
# name or the intended use of these models
name: "split_proj"
version: "1.0.0"
config-version: 2

# This setting configures which "profile" dbt uses for this project.
profile: "split_proj"

# These configurations specify where dbt should look for different types of files.
# The `model-paths` config, for example, states that models in this project can be
# found in the "models/" directory. You probably won't need to change these!
model-paths: ["models"]
analysis-paths: ["analyses"]
test-paths: ["tests"]
seed-paths: ["seeds", "jaffle_data"]
# seed-paths: ["jaffle_data"]
macro-paths: ["macros"]
snapshot-paths: ["snapshots"]

require-dbt-version: ">=1.6.0"

clean-targets: # directories to be removed by `dbt clean`
  - "target"
  - "dbt_packages"

vars:
  truncate_timespan_to: "{{ current_timestamp() }}"

# Configuring models
# Full documentation: https://docs.getdbt.com/docs/configuring-models

# In this example config, we tell dbt to build all models in the example/
# directory as views. These settings can be overridden in the individual model
# files using the `{{ config(...) }}` macro.
models:
  split_proj:
    # Config indicated by + and applies to all files under models/example/
    +on_schema_change: append_new_columns
    example:
      +materialized: view

seeds:
  +schema: jaffle_raw
# the below config for seeds is only relevant when seeding the project for the fist time with the raw jaffle data
# pont the seed path to the jaffle data directory and make sure this is uncommented

# seeds:
#   split_proj:
#     +schema: jaffle_raw

```

## Command

```console
dbt-meshify --debug split orders --select +stg_orders+           
```

## Output
```yml
name: orders
config-version: 2
model-paths:
  - models
macro-paths:
  - macros
seed-paths:
  - seeds
  - jaffle_data
test-paths:
  - tests
analysis-paths:
  - analyses
snapshot-paths:
  - snapshots
clean-targets:
  - target
  - dbt_packages
profile: split_proj
require-dbt-version: '>=1.6.0'
models:
  orders:
    +on_schema_change: append_new_columns
    example:
      +materialized: view
seeds:
  +schema: jaffle_raw
vars:
  truncate_timespan_to: '{{ current_timestamp() }}'

```

### DIFF
--- a/dbt_meshify/storage/dbt_project_editors.py
+++ b/dbt_meshify/storage/dbt_project_editors.py
@@ -96,9 +96,13 @@ class DbtSubprojectCreator:
         contents.pop("query-comment")
         contents = filter_empty_dict_items(contents)
 
-        # Serialize the `require-dbt-version` field into a string
+        # Serialize the `require-dbt-version` field into a string. This happens because dbt is expecting a list of
+        # strings, but also accepts a singular string value, too. We can handle the latter case by checking the length
+        # of each item in the list. If they're all one character long, then we're really working with a single version
+        # string.
         if "require-dbt-version" in contents:
-            contents["require-dbt-version"] = "".join(contents["require-dbt-version"])
+            if max([len(version) for version in contents["require-dbt-version"]]) == 1:
+                contents["require-dbt-version"] = "".join(contents["require-dbt-version"])
 
         return FileChange(
             operation=Operation.Add,
--- a/dbt_meshify/storage/dbt_project_editors.py
+++ b/dbt_meshify/storage/dbt_project_editors.py
@@ -95,7 +95,11 @@ class DbtSubprojectCreator:
         # this one appears in the project yml, but i don't think it should be written
         contents.pop("query-comment")
         contents = filter_empty_dict_items(contents)
-        # project_file_path = self.target_directory / Path("dbt_project.yml")
+
+        # Serialize the `require-dbt-version` field into a string
+        if "require-dbt-version" in contents:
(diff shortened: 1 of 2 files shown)


dave-connors-3:
This looks good! I think we'd also want to test out the case when someone has a range of dbt versions in their project:

```yml
require-dbt-version: [">=1.6.0", "<1.7.0"]
```

I wonder if we'd accidentally end up with a big long string with this approach!

nicholasyager:
> This looks good! I think we'd also want to test out the case when someone has a range of dbt versions in their project:
> 
> ```yaml
> require-dbt-version: [">=1.6.0", "<1.7.0"]
> ```
> 
> I wonder if we'd accidentally end up with a big long string with this approach!

Ha! that'd definitely break. Pulling this back.

nicholasyager:
Alrighty, this has been update to support both singular version strings and lists of versions.

dave-connors-3 (approved):
🦞

PR #155, opened 2023-08-25
check child package

### LINKED ISSUE #154: split errors when a model has children from a package
### Describe the bug

When splitting, we chack the child map for which resources we should go and edit with cross project refs. If that child map includes a model from another package, we will try to edit the code in that resource erroneously, looking in the wrong path:

```
    "model.fishtown_internal_analytics.stg_snowplow_web_page_contexts": [
      "model.fishtown_internal_analytics.stg_dbt_cloud_ide_events",
      "model.snowplow.snowplow_base_web_page_context"
    ],
```

### Steps to reproduce

Have a model as the parent of a package model, try to split it out. 

### Expected results

We skip package models

### Actual results

we try to read the model from the parent project, where it does not exist, throwing a `FileNotFoundError`

dave-connors-3:
resolves #154

worked for my split on internal analytics!

### DIFF
--- a/dbt_meshify/change_set_processor.py
+++ b/dbt_meshify/change_set_processor.py
@@ -27,7 +27,6 @@ class ChangeSetProcessor:
         file_editor = (
             RawFileEditor() if change.entity_type == EntityType.Code else ResourceFileEditor()
         )
-
         file_editor.__getattribute__(change.operation)(change)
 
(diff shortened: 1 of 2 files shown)


nicholasyager (approved):
@dave-connors-3 This PR is beautifully short and sweet. Thanks for making this fix 🪨

Issue #87, opened 2023-06-30
Running the command when no `dbt_project.yml` is available crashes


b-per:
### Describe the bug

```
Runtime Error
  No dbt_project.yml found at expected path /mypath/dbt-meshify/dbt_project.yml
  Verify that each entry within packages.yml (and their transitive dependencies) contains a file named dbt_project.yml
  
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/mypath/cli.py", line 74, in wrapper_decorator
    return func(*args, **kwargs)
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/mypath/main.py", line 237, in group
    ctx.forward(create_group)
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 781, in forward
    return __self.invoke(__cmd, *args, **kwargs)
  File "/mypath/dbt-meshify/.venv/lib/python3.9/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/mypath/cli.py", line 74, in wrapper_decorator
    return func(*args, **kwargs)
  File "/mypath/main.py", line 182, in create_group
    project = DbtProject.from_directory(path)
  File "/mypath/dbt_projects.py", line 172, in from_directory
    manifest=dbt.parse(directory),
  File "/mypath/dbt.py", line 28, in parse
    return self.invoke(directory, ["--quiet", "parse"])
[...]

PR #163, opened 2023-09-05
Fix: Raise a DbtMeshifyError if a project path is provided that does not contain a `dbt_project.yml` file

nicholasyager:
# Description and Motivation
Previously, it was possible for a user to provide a project path that is not real. For instance, running the command in the wrong directory! In this PR, I've added a check to include that a `dbt_project.yml` file is present in the project path, and if not, a `DbtMeshifyException` is raised with a corresponding error message. Along the way, I added tests for each command to confirm that the correct error message is returned if an invalid project path is provided.

Resolves: #87 

# Example

```console
(dbt-meshify-py3.11) > $ dbt-meshify connect --project-paths test-projects/temp_proj/src_proj_a  test-projects/temp_proj/src_proj_b
19:50:17 | ERROR | The provided directory (/Users/nicholas/projects/nicholasyager/dbt-meshify/test-projects/temp_proj/src_proj_a) does not contain a dbt project.
```

### DIFF
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -25,6 +25,7 @@ from dbt.node_types import NodeType
 from loguru import logger
 
 from dbt_meshify.dbt import Dbt
+from dbt_meshify.exceptions import FatalMeshifyException
 
 
 class BaseDbtProject:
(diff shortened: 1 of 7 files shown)


dave-connors-3 (approved):
🦞

PR #164, opened 2023-09-06
add --force flag to deploy steps

dave-connors-3:
per @b-per, there's a [known issue](https://github.com/mkdocs/mkdocs/issues/2370) related to the error message we saw in yesterday's docs deploy job. per this thread, adding the `--force` flag should likely resolve our issue.


nicholasyager (approved):
This seems straightforward to me! ✅

PR #167, opened 2023-09-06
Update examples.md

graciegoheen:


docs/examples.md
```diff
@@ -1,6 +1,6 @@
 # Examples
 
-For consistency and clarity of the following examples, we're going to use a simplified dbt project. In practice, the model governance features describe are _most_ beneficial for large dbt projects that are struggling to scale.
+For consistency and clarity of the following examples, we're going to use a simplified dbt project. In practice, the dbt mesh features describe are _most_ beneficial for large dbt projects that are struggling to scale.
```
nicholasyager, line 3: I'd really like to avoid using the term "dbt mesh". I have two reasons:
1. Right now, it doesn't mean anything to someone outside dbt Labs. Most of the industry is familiar with the idea of data mesh and the idea of data governance, so it would behoove us to use these widely-adopted terms. My vote is currently for "data mesh".

    > In practice, the _data mesh_ features described are most beneficial for large dbt projects ...
2. This is more personal, but I have a hunch that "dbt mesh" is a good candidate to be trademarked by dbt Labs to represent the proprietary multi-project collaboration product within dbt Cloud. If this were to occur, I wouldn't want this open-source project to tied to dbt Labs' proprietary offering as the primary use-case -- particularly when it provides so much value for dbt-core users.

graciegoheen: Also just a note to self that if we update this: will need to make it consistent in all the docs for this package

nicholasyager: > model governance and cross-project lineage
This makes sense to me! 👍🏻

graciegoheen: updated!

nicholasyager (approved):
@graciegoheen I appreciate the consistency that this PR creates in our docs. Thank you for making this ✅

PR #171, opened 2023-09-11
default prerelease to true

### LINKED ISSUE #169: `--pre-release` should be the default behavior
### Describe the feature

It seems to be reasonable that most of the time you're incrementing model versions, you _won't increment the model's `latest_version`_ -- there will generally be a period of transition where you'll want to ensure new changes are non-breaking, and consumer have time to update their logic. The pre-release flag controls that behavior currently, but it's defaulted to false. I think we should update this behavior!

Do nothing, eat more sunchips

### Who will this benefit?

Model versioners of all stars and stripes

Sure thing pal

dave-connors-3:
Closes #169 

This PR inverts the `--pre-release` flag, and replaces it with `--no-prerelease` -- the thought here is that when versioning, the default path would be to add a new version, but _not_ increment latest version to give your consumers time to upgade. 

I was a bit lazy with the refactor to just invert the flag value and pass that to the `prerelease` argument we have in the versioning logic. If we'd prefer to update all references, I can do that too. 

If we like this approach, I will update the docs too!


dave-connors-3:
todo: docs + rename flag to `--latest`

nicholasyager (approved):
@dave-connors-3 This is a great quality of life improvement! Special kudos for reversing the logic (`--no-prerelease` -> `--latest`).

Let's do it! 🚀 ✅

dave-connors-3:
gonna start fresh on a new branch

PR #173, opened 2023-09-16
handle version args for ref updates in sql and python

### LINKED ISSUE #172: Ref updater does not account for version argument
we don't handle the case when refs contain a version argument in `references.ReferenceUpdater`! we shoudl

dave-connors-3:
This PR properly handles versioned refs when performing the regex replace to add the project argument during splitting. 

Closes #172

### DIFF
--- a/dbt_meshify/utilities/references.py
+++ b/dbt_meshify/utilities/references.py
@@ -48,13 +48,20 @@ class ReferenceUpdater:
         """Update refs in a SQL file."""
 
         # pattern to search for ref() with optional spaces and either single or double quotes
-        pattern = re.compile(r"{{\s*ref\s*\(\s*['\"]" + re.escape(model_name) + r"['\"]\s*\)\s*}}")
+        # this also captures the version argument if it exists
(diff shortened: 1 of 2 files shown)


dbt_meshify/utilities/references.py
```diff
@@ -48,13 +48,20 @@ def update_refs__sql(model_name, project_name, model_code):
...
+        )
 
-        # replacement string with the new format
-        replacement = f"{{{{ ref('{project_name}', '{model_name}') }}}}"
+        def replacement_function(match):
+            version_arg = match.group(1)
+            if version_arg:
+                return f"{{{{ ref('{project_name}', '{model_name}', {version_arg}) }}}}"
+            else:
+                return f"{{{{ ref('{project_name}', '{model_name}') }}}}"
```
nicholasyager, line 61: I really like this functiob-based approach. 👍🏻

nicholasyager (approved):
This Pr seems both necessary and sufficient for enabling versioned ref updates. Kudos on the regex! ✅

Issue #168, opened 2023-09-11
first `version` operation should result in 2 versions


dave-connors-3:
### Describe the feature

Right now, for an unversioned model, running `dbt-meshify operation add-version -s my_model` results in just one version of the model -- it effectively promotes the unversioned model to the versioning paradigm, but doesn't actually increment/update the version of the model. Instead, we should add an additional versioned file so that the first invocation results in 2 versions. 

### Describe alternatives you've considered

Do nothing, eat some sunchips

### Are you interested in contributing this feature?

Yes I am

dave-connors-3:
after discussion:

1. break `add-version` and `create-version` into two separate `operation` commands
2. add new top level `version` command that does both

PR #175, opened 2023-09-16
Feature: Separate the `add-version` command into `add-version` and `bump-version`. Create new `version` command.

nicholasyager:
# Description of Changes
This PR makes some lofty changes to how versions are handled in dbt-meshify.
1. The `operation add-version` command ONLY adds version boilerplate for models that are currently unversioned.
2. A new `operation bump-version` command handles bumping versions for models, including prereleases and file operations.
3. A new `version` command, which does both! In this case, previously unversioned models will have version boilerplate added, and then the version will be incremented on top.

Resolves: #168

### DIFF
--- a/dbt_meshify/utilities/versioner.py
+++ b/dbt_meshify/utilities/versioner.py
@@ -2,6 +2,7 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 from dbt.contracts.graph.nodes import ModelNode
+from loguru import logger
 
 from dbt_meshify.change import (
     ChangeSet,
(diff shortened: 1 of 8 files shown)


dbt_meshify/utilities/versioner.py
```diff
@@ -58,11 +62,11 @@ def get_latest_model_version(model_versions: NamedList) -> int:
         except ValueError:
             raise ValueError("Version not an integer, can't increment version.")
 
-    def generate_version(
-        self, model: ModelNode, prerelease: bool = False, defined_in: Optional[Path] = None
-    ) -> ChangeSet:
-        """Create a Change that adds a Version config to a Model."""
+    def add_version(self, model: ModelNode, defined_in: Optional[Path] = None):
```
dave-connors-3, line 65: not a big deal but it looks like we lost the return typehint for this method!

dave-connors-3 (approved):
great work here! i like this interface quite a bit, thanks for the quick turnaroun!

dbt_meshify/utilities/versioner.py
nicholasyager: Ugh. Gotta add that back.

Issue #177, opened 2023-09-19
new `version` command raises a `FatalMeshifyError`


dave-connors-3:
### Describe the bug

the new `version` command seems to run into an error when resolving the `FileChange` for copying code 

### Steps to reproduce

1. navigate to our `src_proj_a`
2. run `dbt-meshify version -s shared_model` 
3. see `FatalMeshifyException` rasied

```
16:42:59 | INFO | Reading dbt project at /Users/daveconnors/dev/dbt-labs/dbt-meshify/test-projects/source-hack/src_proj_a
16:42:59 | INFO | Executing dbt parse...
16:43:00 | INFO | Generating catalog with dbt docs generate...
16:43:01 | INFO | Selected 1 resources: ['model.src_proj_a.shared_model']
16:43:01 | INFO | Adding version to models in selected resources...
16:43:01 | [  1/3] STARTING | Move code `shared_model` to models/shared_model_v1.sql
16:43:01 | [  1/3] SUCCESS  | Move code `shared_model` to models/shared_model_v1.sql
16:43:01 | [  2/3] STARTING | Add model `shared_model` to models/_models.yml
16:43:01 | [  2/3] SUCCESS  | Add model `shared_model` to models/_models.yml
16:43:01 | [  3/3] STARTING | Copy code `shared_model` to models/shared_model_v1.sql
16:43:01 | ERROR | PosixPath('/Users/daveconnors/dev/dbt-labs/dbt-meshify/test-projects/source-hack/src_proj_a/models/shared_model_v1.sql') and PosixPath('/Users/daveconnors/dev/dbt-labs/dbt-meshify/test-projects/source-hack/src_proj_a/models/shared_model_v1.sql') are the same file
```

<details>
<summary> Full Logs</summary>

```
❯ dbt-meshify version -s shared_model
16:42:59 | INFO | Reading dbt project at /Users/daveconnors/dev/dbt-labs/dbt-meshify/test-projects/source-hack/src_proj_a
16:42:59 | INFO | Executing dbt parse...
16:43:00 | INFO | Generating catalog with dbt docs generate...
16:43:01 | INFO | Selected 1 resources: ['model.src_proj_a.shared_model']
16:43:01 | INFO | Adding version to models in selected resources...
16:43:01 | [  1/3] STARTING | Move code `shared_model` to models/shared_model_v1.sql
16:43:01 | [  1/3] SUCCESS  | Move code `shared_model` to models/shared_model_v1.sql
[...]

PR #178, opened 2023-09-20
Fix: Correct behavior of `version` for models missing YAML properties

nicholasyager:
# Description
In issue #177, it's observed that a `version` run failed due to an attempt to Copy a file to the same location as the source. This is a result of two bits of code interacting in unexpected ways:
1. The `load_model_yaml` file was failing due to the model being versioned not having YAML.
2. The `model_override` not being applied to the loaded model yaml, due to this logic being placed within the `try/except` block.

As a result, if a model missing YAML properties was being versioned, the output from `add_version` would not be added to `bump_version`, and it would fail.

Outside remedying this issue, there are two other systematic failures in place, too:
1. `bump_version` should not support un-versioned models. A new exception has been added to handle this, and tests have been updated accordingly.
2. We were missing _basic_ tests for `version`. To that end, I've added a _baseline_ integration test to ensure that the dang command actually works right.
   - On this note, @dave-connors-3, please let me know if there are other test cases you'd like me to cover for this meta command that are not covered within the operation commands.

Resolves: #177

### DIFF
--- a/dbt_meshify/utilities/versioner.py
+++ b/dbt_meshify/utilities/versioner.py
@@ -165,7 +165,7 @@ class ModelVersioner:
         next_version_file_name = model_path.parent / Path(
             f"{defined_in}.{model.language}"
             if defined_in
-            else f"{model.name}_v{new_latest_version_number}.{model.language}"
+            else f"{model.name}_v{new_greatest_version_number}.{model.language}"
         )
 
         change_set = ChangeSet()
--- a/dbt_meshify/utilities/versioner.py
+++ b/dbt_meshify/utilities/versioner.py
@@ -2,6 +2,7 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 from dbt.contracts.graph.nodes import ModelNode
+from loguru import logger
 
 from dbt_meshify.change import (
     ChangeSet,
@@ -32,7 +33,11 @@ class ModelVersioner:
 
     def load_model_yml(self, path: Path, name: str) -> Dict:
         """Load the Model patch YAML for a given ModelNode."""
-        raw_yml = self.file_manager.read_file(path)
+        try:
+            raw_yml = self.file_manager.read_file(path)
+        except FileNotFoundError as e:
+            logger.debug(f"Unable to load file for versioning. {path}. {e}")
+            return {}
 
         if not isinstance(raw_yml, dict):
             raise Exception(
@@ -68,10 +73,7 @@ class ModelVersioner:
         path = self.project.resolve_patch_path(model)
         model_path = self.project.path / model.original_file_path
 
-        try:
-            model_yml = self.load_model_yml(path, model.name)
-        except FileNotFoundError:
-            model_yml = {}
+        model_yml = self.load_model_yml(path, model.name)
 
         model_versions: NamedList = self.get_model_versions(model_yml)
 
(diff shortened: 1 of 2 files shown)


dave-connors-3:
This looks a lot better Nick! I was able to find the same error as described in the issue when using the prerelease flag on these changes:

1. `dbt-meshify operation add-version -s shared_model` resulted in the expected yml (latest = 1, versions = [1])
2.`dbt-meshify operation bump-version -s shared_model --prerelease` resulted in the expected yml (latest = 1, versions = [1]), but hit the same error when copying code from v_1 to v_2

nicholasyager:
🤔 Ya know, I bet we're running into a version number conflict with
pre-release. Specifically if `latest_version` + 1 already exists. Might
need to tweak this to use `greatest_version` + 1 instead. Definitely need
more tests for this.

Thanks for the feedback!

dbt_meshify/utilities/versioner.py
```diff
@@ -134,19 +136,21 @@ def bump_version(
...
+            raise ModelVersionerException(
+                f"The model {model.name} does not have any versions defined. Please use add-version first."
+            )
+
+        # Within dbt-core, if unset `latest_version` defaults to the greatest version identifier.
+        latest_version = model_yml.get("latest_version", greatest_version)
+
         # Bump versions
         new_greatest_version_number = greatest_version + 1
         new_latest_version_number = latest_version if prerelease else latest_version + 1
```
dave-connors-3, line 156: I think this line may be the source of the error -- when bumping with prerelease from this yml:

```yml
models:
  - name: shared_model
    latest_version: 1
    description: "this is a test model"
    versions:
      - v: 1
```

we'd expect the `latest_version` to remain `1` but to have a new file at `shared_model_v2`. 

the filepath calc a few lines down is 

```
next_version_file_name = model_path.parent / Path(
            f"{defined_in}.{model.language}"
            if defined_in
            else f"{model.name}_v{new_latest_version_number}.{model.language}"
        )
```

which in this case would be `shared_model_v1.sql` which already exists! I think it's only at this first point that this is an issue, so we might just need to add custom logic for when this is your starting point?

dave-connors-3: the fix might be as simple as using `new_greatest_version_number`  for the filepath calculation?

nicholasyager: 100000% agreed

nicholasyager:
@dave-connors-3 Thanks for the investigation! I was able to replicate the issue via pytest and your solution resolves the issue.

dave-connors-3:
@nicholasyager after some local testing, was able to confirm this behaves as expected:

| dbt-meshify command    | model without yml or versions | unversioned model with yml | versioned model                 |
|------------------------|-------------------------------|----------------------------|---------------------------------|
| operation bump-version | Error, no action              | Error, no action           | +1 version, latest = latest + 1 |
| operation add-version  | Add v1, v1=latest             | Add v1, v1=latest          | Error, no action                |
| version                | add v1, v2, v2 = latest       | add v1, v2, v2 = latest    | +1 version, latest = latest + 1 |

once this is in, and I update the flags, I think we'll have exactly the versioning behavior we were after!

tests/integration/test_version_command.py
```diff
@@ -299,3 +388,16 @@ def test_command_raises_exception_invalid_paths():
 
     assert result.exit_code != 0
     assert "does not contain a dbt project" in result.stdout
+
+
+def test_version_no_model_yaml(project):
+    """Verify that the version command running in default mode adds versions and creates a second version, too"""
```
dave-connors-3, line 394: can you add an assertion that we get two versions of the model here? and/or update the comment to say "just test that the command runs in default mode"

dave-connors-3 (approved):
small comment on the `dbt-meshify` version command test, but otherwise ready to go!

tests/integration/test_version_command.py
nicholasyager: Fantastic call. Will add.

dave-connors-3 (approved):
LGTM! (🦞 good to me)

PR #179, opened 2023-09-24
default prerelease to true

### LINKED ISSUE #169: `--pre-release` should be the default behavior

dave-connors-3:
Closes #169

This PR is a re-do of closed #171 so that I don't need to deal with merge conflicts on a Sunday!

This reverses the default behavior for the version commands, in particular, the default argument of `prerelease` to the `ModelVersioner.bump_version()` method, as controlled by the `--latest` flag -- default behavior is to add a new version, but not increment the `latest_version` unless the `--latest` flag is present


nicholasyager (approved):
Given that this is simply updated from the previous `latest` version PR, this is likely still good! ✅

PR #182, opened 2023-09-28
pass `--read-catalog` to both `group` subcommands

### LINKED ISSUE #181: when using `group` `--read-catalog` only applied to first of 2 sub commands
### Describe the bug

In `dbt-meshify group` we execute a `operation create-group` and then an `add-contract`, but only pass the `--read-catalog` to the first command, so we end up running docs generate either way!

dave-connors-3:
closes #181

Makes sure we properly pass the read catalog flag to the second invocation nested in `dbt-meshify group` !

### DIFF
--- a/dbt_meshify/main.py
+++ b/dbt_meshify/main.py
@@ -629,6 +629,7 @@ def group(
             ]
         ),
         project_path=project_path,
+        read_catalog=read_catalog,
     )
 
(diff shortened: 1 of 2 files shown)


nicholasyager (approved):
@dave-connors-3 I appreciate how straightforward the fix here is. Kudos for adding the test. ✅ 🚀

PR #183, opened 2023-09-29
176 clean before deps

### LINKED ISSUE #176: Run `dbt clean` before `dbt deps`
### Describe the feature
#174 adds the ability to run `dbt deps` automatically when packages are missing. There might be cases when `dbt_packages.yml` is modified and we removed some packages. In that case, it could be useful to run `dbt clean` beforehand to remove older packages no longer required. 

Of course!

b-per:
Fixes #176 

Runs `dbt clean` before `dbt deps` when the packages installed don't match the packages in `packages.yml`

One of the use cases where clean is required for example is when people used to use `dbt_metrics` and then move to 1.6. Removing the package from `packages.yml` doesn't remove it from `dbt_packages` and still raise compilation errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ venv/
 __pycache__/
 # IDEs
 .idea/
+.vscode/
 
 **/database.db
 logs/
--- a/dbt_meshify/dbt.py
+++ b/dbt_meshify/dbt.py
@@ -22,7 +22,8 @@ class Dbt:
         result = self.dbt_runner.invoke(runner_args if runner_args else [])
         if not result.success and result.exception:
             if isinstance(result.exception, UninstalledPackagesFoundError):
-                logger.debug("Project missing packages, running dbt deps...")
+                logger.debug("Project missing packages, running dbt clean and deps...")
+                self.dbt_runner.invoke(["clean"])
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ venv/
 __pycache__/
 # IDEs
 .idea/
+.vscode/
```
nicholasyager, line 10: I'm genuinely surprised we've managed to get this far without adding `.vscode` to the gitignore. Thanks for adding it here.

dbt_meshify/dbt.py
```diff
@@ -22,7 +22,8 @@ def invoke(
         result = self.dbt_runner.invoke(runner_args if runner_args else [])
         if not result.success and result.exception:
             if isinstance(result.exception, UninstalledPackagesFoundError):
-                logger.debug("Project missing packages, running dbt deps...")
+                logger.debug("Project missing packages, running dbt clean and dbt deps...")
```
nicholasyager, line 25: Kudos for updating the debug log.

nicholasyager (approved):
@b-per This change is both necessary and sufficient for resolving #176. Thanks for the contribution.

PR #185, opened 2023-10-09
pass args as string

dave-connors-3:
This PR makes dbt-meshify compatible with the latest dbt core release by passing any multiselect args to `dbt ls` as a single string separated by spaces, which is accepted by the parser in dbt core. 

My initial implementation was to convert the tuple that is returned from our meshify command to a list, but that raised the same error as before -- if appears the type check we were discussing in our sync is not relevant for the `--select` flag, and is used for other flags. 

I tested on 1.6.3+, and these changes appear to work across versions


nicholasyager (approved):
@dave-connors-3 I have one small suggestion to remove a debugger line, but otherwise this looks good!

Issue #158, opened 2023-09-05
Check if split command copies over files with doc blocks


graciegoheen:
🤔 Some low risk thoughts
- `split` doesn't copy over all project files, so there may be documentation that's not copied over (like markdown files).
  - Now that I'm thinking about it, does split copy over files with docblocks?
- `group` and `create-group` will update existing groups already defined in yml.

_Originally posted by @nicholasyager in https://github.com/dbt-labs/dbt-meshify/pull/115#discussion_r1312985891_

dave-connors-3:
did a little poking at docs blocks:

- models that depend on docs blocks move just fine
- there's not any evidence in the manifest that i can see that ties a given `manifest.docs` entry with `manifest.nodes` entries -- i am quite curious how core currently resolves and compiles docs blocks! i think until we know how that relationship is established, we won't be able to correctly move docs blocks unless we did something hacky like copy over any docs blocks that share paths with the other selected files (assuming people follow structure best practices)

Issue #142, opened 2023-08-21
Copying macro code copies the full contents of the sql file, including other macros


dave-connors-3:
### Describe the bug

```
in macros/foo_and_bar.sql

{% macro foo() %}
{% endmacro %}

{% macro bar() %}
{% endmacro %}
```

if you select a model that uses `bar`, both macros are copied over since we don't parse the SQL

### System information
**Which database are you using dbt with?**
- [ ] postgres
- [ ] redshift
- [ ] bigquery
- [ ] snowflake
- [ ] other (specify: ____________)

**The output of `dbt debug`:**
```
<output goes here>
```

**The output of `dbt --version`:**
```
<output goes here>
```

PR #192, opened 2023-12-29
Feature: Block-level copying of Jinja content between split projects

nicholasyager:
# Description and Motivation

There are two PRs related to copying Jinja blocks to split dbt projects. One has to do with block-level copying of macros, to prevent multiple macros in the same file being copyied to a split project, and the other has to do with doc blocks not being copied over due to an absence of references. This PR resolves both issues.

1. This PR creates a new data structure for keeping track of `JinjaBlocks` at a character position level. This is accomplished by using `docs` and `macros` properties of a `Manifest`, coupled with some clever RegEx, to find where things start and end. Note that the RegEx may be brittle, so 👀 
2. Adds an `append` Operation to our file editors, allowing us to append-only edit files (like if we're moving a Jinja block from one place to another!).
3. Adds reference tracking code to `DbtSubProject` so we know which doc blocks are being referenced in which code and patch files. This allows us to track which Doc blocks need to be copied, too!
4. Add new tests and update existing tests!

Resolves: #142
Resolves: #158

### DIFF
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -320,8 +320,19 @@ class DbtProject(BaseDbtProject, PathedProject):
         blocks = {}
 
         for unique_id, item in self.manifest.docs.items():
+            if item.package_name != self.name:
+                continue
+
             blocks[unique_id] = JinjaBlock.from_file(
                 path=self.path / item.original_file_path, block_type="docs", name=item.name
+            )
+
+        for unique_id, macro in self.manifest.macros.items():
+            if macro.package_name != self.name:
+                continue
+
+            blocks[unique_id] = JinjaBlock.from_file(
+                path=self.path / macro.original_file_path, block_type="macro", name=macro.name
             )
 
         return blocks
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -311,7 +311,6 @@ class DbtProject(BaseDbtProject, PathedProject):
         resources = self.select_resources(output_key="unique_id")
 
         super().__init__(manifest, project, catalog, name, resources)
-
         self.jinja_blocks: Dict[str, JinjaBlock] = self.find_jinja_blocks()
 
     def find_jinja_blocks(self) -> Dict[str, JinjaBlock]:
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -474,7 +474,7 @@ class DbtSubProject(BaseDbtProject, PathedProject):
             node = self.get_manifest_node(unique_id)
 
             if node is None:
-                continue
+                raise Exception(f"Unable to find referenced node {node}")
 
             if hasattr(node, "raw_code"):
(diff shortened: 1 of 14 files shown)


dbt_meshify/storage/jinja_blocks.py
```diff
@@ -0,0 +1,75 @@
...
+    @staticmethod
+    def isolate_content(file_content: str, start: int, end: int) -> str:
+        """Given content, a start position, and an end position, return the content of a Jinja block."""
+        return file_content[start:end]
+
+    @classmethod
+    def from_file(cls, path: Path, block_type: str, name: str) -> "JinjaBlock":
+        """Find a specific Jinja block within a file, based on the block type and the name."""
+
+        file_content = open(path).read()
```
dave-connors-3, line 62: is this different that `Path.read_text()`? i think we've used that method elsewhere

dbt_meshify/dbt_projects.py
```diff
@@ -432,6 +459,39 @@ def _get_custom_macros(self) -> Set[str]:
...
+            return self._referenced_docs
+
+        docs = set()
+        for unique_id in self.resources:
+            if unique_id.startswith("test."):
+                continue
+
+            node = self.get_manifest_node(unique_id)
+
+            if node is None:
```
dave-connors-3, line 476: should this throw an error? i think we throw a similar one if we can't find a node in the manifest elsewhere

dbt_meshify/dbt_projects.py
```diff
@@ -432,6 +459,39 @@ def _get_custom_macros(self) -> Set[str]:
...
+                continue
+
+            if hasattr(node, "raw_code"):
```
dave-connors-3, line 479: forgetting which nodes have `raw_code` -- does this indicate we'd be searching the content of a model for docs blocks? are we looking for docs in the definition of YML only resources in this step?

dbt_meshify/dbt_projects.py
```diff
@@ -432,6 +459,39 @@ def _get_custom_macros(self) -> Set[str]:
...
+                docs.update(find_doc_reference(node.raw_code))
+
+            if hasattr(node, "patch_path"):
```
dave-connors-3, line 482: ah i think this answers my above question

test-projects/split/split_proj/models/marts/__models.yml
```diff
@@ -53,7 +58,8 @@ models:
               to: ref('stg_customers')
               field: customer_id
       - name: location_id
-        description: The foreign key relating to the location the order was placed
+        description:
```
dave-connors-3, line 61: a note for future us -- we gotta see if ruamel can stop editing newlines 😅

dave-connors-3 (approved):
i dig it! nice work on this

dbt_meshify/dbt_projects.py
nicholasyager: I think you've figured this out already, but yes. This checks `raw_code`, and the next check on line 482 checks wihtin the YAML file, too.

PR #194, opened 2024-01-15
update refs in exposure and semantic model definitions when downstream of a split resource

### LINKED ISSUE #191: exposures do not play nice with meshify
### Describe the bug

Report from a customer that exposures "break with meshify" -- this ticket is a stub to investigate. Hunch is that this is on the `split` operation

dave-connors-3:
may close #191 

this is a tricky one!

1. Updated the logic in `update_child_refs()` to update `depends_on` fields for exposures to take a two-arg ref if downstream of a model being split - worked well!
2. Wanted to extend the same logic to semantic models as well -- this is where we took a turn:

- dbt-core did not add the ability to return semantic models in an `ls` function until 1.7+ -- we need this for a variety of operations, so i updated `dbt-core` for this package to `1.7.4`
- after updating, the logic in `dbt ls -s +orders` _includes the semantic models attached to orders_ which means that `dbt-meshify split test --select +orders` moves the orders semantic model th##at I added to the new project. This is a) good that we do what dbt-core tells us to do but b) not at all the behavior that I expected. As such, I added a test that is just a pass until we can figure out what the expected behavior is. 

#### update 1/16 

the reason that the semantic model was being selected and moved was because the model and semantic model had the same name! so this was in fact selected as well. This is expected behavior, but is a bit of a stumbling block if trying to have a semantic model reference a newly split resource. 

That said, i believe this is ready to be reviewed -- the string matching on refs is not my cleanest code so very open to suggestions!

### DIFF
--- a/dbt_meshify/utilities/references.py
+++ b/dbt_meshify/utilities/references.py
@@ -182,30 +182,29 @@ class ReferenceUpdater:
                 model_code=code,
             )
 
-            change = FileChange(
+            return FileChange(
                 operation=Operation.Update,
                 entity_type=EntityType.Code,
                 identifier=downstream_node.name,
                 path=downstream_project.resolve_file_path(downstream_node),
                 data=updated_code,
             )
-            return change
 
-        if isinstance(downstream_node, Exposure) or isinstance(downstream_node, SemanticModel):
+        elif isinstance(downstream_node, Exposure) or isinstance(downstream_node, SemanticModel):
             is_exposure = isinstance(downstream_node, Exposure)
             data = self.update_yml_resource_references(
                 project_name=project_name,
                 upstream_resource_name=upstream_node.name,
                 resource=downstream_node,
             )
-            change = ResourceChange(
+            return ResourceChange(
                 operation=Operation.Update,
                 entity_type=EntityType.Exposure if is_exposure else EntityType.SemanticModel,
                 identifier=downstream_node.name,
                 path=downstream_project.resolve_file_path(downstream_node),
                 data=data,
             )
-        return change
+        raise Exception("Invalid node type provided to generate_reference_update.")
 
     def update_child_refs(
         self,


dbt_meshify/utilities/references.py
```diff
@@ -135,32 +141,77 @@ def replace_source_with_ref__python(
...
+        ref_to_update = f"ref('{upstream_resource_name}')"
+        str_refs = []
+        for ref in refs:
+            package_clause = f"'{ref.package}', " if ref.package else ""
+            name_clause = f"'{ref.name}'"
+            version_clause = f", v={ref.version}" if ref.version else ""
+            str_ref = f"ref({package_clause}{name_clause}{version_clause})"
+            if str_ref != ref_to_update:
+                str_refs.append(str_ref)
+        str_refs.append(new_ref)
```
nicholasyager, line 164: Am I correct in believing that this code fully specifies each resource ref, and then adds the updated upstream source ref? If so, this seems sufficient. Perhaps it would be valuable to create a function that generates ref strings for resources? Either way, ✅

nicholasyager (approved):
@dave-connors-3 This is great! I had a couple of nitpicks in `references.py` around how one might simplify, but those are non-blocking IMHO. This change seems both necessary and sufficient

dbt_meshify/utilities/references.py
```diff
@@ -182,31 +182,29 @@ def generate_reference_update(
...
+                resource=downstream_node,
+            )
+            return ResourceChange(
+                operation=Operation.Update,
+                entity_type=EntityType.Exposure if is_exposure else EntityType.SemanticModel,
+                identifier=downstream_node.name,
+                path=downstream_project.resolve_file_path(downstream_node),
+                data=data,
+            )
+        raise Exception("Invalid node type provided to generate_reference_update.")
```
nicholasyager, line 207: Fantastic Doctor Who GIF

nicholasyager (approved):
@dave-connors-3 Thanks for making some formatting/simplification tweaks. This PR looks both necessary and sufficient ✅ 🚀

PR #198, opened 2024-03-22
Fix example command for adding a version

dbeatty10:
### Version

```shell
$ pip show dbt-meshify 
Name: dbt-meshify
Version: 0.5.2
```

### Current behavior

```shell
$ dbt-meshify version --select customers             

Usage: dbt-meshify [OPTIONS] COMMAND [ARGS]...
Try 'dbt-meshify --help' for help.

Error: No such command 'version'.
```

### Expected behavior

```shell
$ dbt-meshify operation add-version --select customers

Add version 1 to the customers model where none existed previously.
```

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -45,7 +45,7 @@ Let's say you want to add a new version to the customers model, which is current
 You can run the following command:
 
 ```bash
-dbt-meshify version --select customers
+dbt-meshify operation add-version --select customers
 ```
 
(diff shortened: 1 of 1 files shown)


dbeatty10:
I think I had an out-dated version of `dbt-meshify` in my virtual environment.

After making _certain_ I had the latest version of dbt-meshify working in my local environment, the example command did indeed work, so I'm closing this issue.

But opened https://github.com/dbt-labs/dbt-meshify/issues/200 instead, 'cause that would have helped me identify the issue with my environment faster.

Issue #200, opened 2024-03-23
Add a `--version` flag


dbeatty10:
### Describe the feature

```
$ dbt-meshify --version
```

or:

```
$ dbt-meshify -v
```

Gives the same output as:

```shell
python -m pip show dbt-meshify
```

### Additional context
Is this feature dbt Core or dbt Cloud specific? Which database(s) is/are relevant? Please include any other relevant context here.

### Who will this benefit?
What kind of use case will this feature be useful for? Please be specific and provide examples, this will help us prioritize properly.

nicholasyager:
Hi @dbeatty10 👋🏻 Quick clarifying question: 

When you say

> Gives the same output as:
> ```
> python -m pip show dbt-meshify
> ```

You are referring to the package version number, correct? Presumably there isn't a desire to be outputting the other package details too 😁

dbeatty10:
I wasn't precise in my original description 😬 

You got it @nicholasyager -- just the package version number.

#201 looks great 🚀

PR #201, opened 2024-03-23
Feature: Add  a `--version` flag to aid in debugging

nicholasyager:
# Description and motivation
As mentioned in #198, there is a distinct need to have a `--version` flag added to `dbt-meshify` to make it easier for users to identify which version of `dbt-meshify` they have installed. This PR adds this functionality, such that running `dbt-meshify --version` will print the current version number from the package information (as stored w/in the internals of python), and exit.

To accomplish this, I added the new option, and then added the `invoke_without_command=True` argument to `click.group()` to allow us to run code in the `cli` group when a command has not been provided. I then manually implemented a default check to print help text if a subcommand has not been provided.

Resolves: #200 

## Examples

```console
$ dbt-meshify --version
0.5.2                 
```

```console
$ dbt-meshify
Usage: dbt-meshify [OPTIONS] COMMAND [ARGS]...

Options:
  --dry-run
  --debug
  --version  Show version information and exit
  --help     Show this message and exit.

Commands:
  connect    Connects multiple dbt projects together by adding all...
  group      Creates a new dbt group based on the selection syntax...
  operation  Set of subcommands for performing mesh operations on dbt...
  split      Splits out a new subproject from a dbt project by adding all...
  version    Increment the models to the next version, and create in the...          
```

### DIFF
--- a/tests/integration/test_cli_group.py
+++ b/tests/integration/test_cli_group.py
@@ -27,5 +27,5 @@ def test_no_command_prints_help():
 
     assert result.exit_code == 0
 
-    help_result = result = runner.invoke(cli, ["--help"])
+    help_result = runner.invoke(cli, ["--help"])
     assert result.stdout == help_result.stdout
--- a/tests/integration/test_cli_group.py
+++ b/tests/integration/test_cli_group.py
@@ -1,0 +1,31 @@
+from click.testing import CliRunner
+
+from dbt_meshify.cli import get_version
+from dbt_meshify.main import cli
+
+
+def test_version_option():
+    """When the --version option is passed, meshify returns the version and exits."""
+
+    runner = CliRunner()
+    result = runner.invoke(
+        cli,
+        [
+            "--version",
+        ],
+    )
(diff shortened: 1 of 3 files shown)


tests/integration/test_cli_group.py
```diff
@@ -0,0 +1,31 @@
...
+
+def test_no_command_prints_help():
+    """When an invoked subcommand is not provided, print the help and exit."""
+
+    runner = CliRunner()
+    result = runner.invoke(cli)
+
+    assert result.exit_code == 0
+
+    help_result = result = runner.invoke(cli, ["--help"])
```
dave-connors-3, line 30: is this an intentional inclusion of `result` on this line?

dave-connors-3 (approved):
pending comment on the test, but looks good otherwise!

tests/integration/test_cli_group.py
nicholasyager: Nope! I'll fix that.

Issue #124, opened 2023-08-18
split should create all relevant dbt project files and folders


dave-connors-3:
Another thing I found odd here, was that only the folders that we're copying over are create. 

In my case, there was no empty folder for tests, seeds, macros, etc. that I'd expect when running a `dbt init`.

<img width="382" alt="Screenshot 2023-08-17 at 5 01 38 PM" src="https://github.com/dbt-labs/dbt-meshify/assets/53586774/4156967a-216a-4288-9b6b-283df8675039">

_Originally posted by @graciegoheen in https://github.com/dbt-labs/dbt-meshify/issues/120#issuecomment-1683133612_

nicholasyager:
Throwing my research notes and ideas in here as they pop up. I've identified the source of dbt-core's default project files in `dbt/include/starter_project`. My intention is to update the `DbtSubprojectCreator.write_project_file()` method to copy over the directory structure from here into the target project directory. Then, copy over a clean `dbt_project.yml` file and update it with the relevant configs from the parent project.

Open questions:
- Will pyyaml.rumel preserve comments?
- We may need to be mindful of the order of operations to prevent duplicate changes from firing (attempting to creating a directory twice)

PR #206, opened 2024-05-05
Feature: Use dbt-core starter project as baseline for split projects

nicholasyager:
# Description and motivation
A common question for users of dbt-meshify is "why does the split project look ... odd?" In most cases, this is a byproduct of the existing `split` command creating only necessary directories and files in the split project. While functional, this is not the ideal experience for users.

This PR updates our method for creating projects, to instead start by creating a new starter project in the target directory, add all directories and files of interest, and then update the starter `dbt_project.yml` file using the sub-project's configuration information. The benefits to this approach are:
1. Sub-projects have all the standard files and directories for a dbt project.
2. The sub-project have a familiar `dbt_project.yml` file experience.

Along the way, I also implemented the following:
1. Tests that confirm the presence of standard dbt directories in sub-projects on creation.
2. Added a new Change type (`Directorychange`), file manager (`DirectoryManager`) and directory editor (`DirectoryEditor`) for interacting with directory structures instead of files. Some minor refactoring was implemented along the way.

Resolves: #124 
Resolves: #153 

## Example `dbt_project.yml`

```yml
# Name your project! Project names should contain only lowercase characters
# and underscores. A good package name should reflect your organization's
# name or the intended use of these models
name: 'customers'
version: '1.0.0'
config-version: 2

# This setting configures which "profile" dbt uses for this project.
profile: 'split_proj'

# These configurations specify where dbt should look for different types of files.
# The `model-paths` config, for example, states that models in this project can be
# found in the "models/" directory. You probably won't need to change these!
model-paths:
  - models
analysis-paths:
  - analyses
test-paths:
  - tests
seed-paths:
  - seeds
  - jaffle_data
macro-paths:
  - macros
snapshot-paths:
  - snapshots

clean-targets:         # directories to be removed by `dbt clean`
  - target
  - dbt_packages
models:
  customers:
    +on_schema_change: append_new_columns
    example:
      +materialized: view
require-dbt-version:
  - '>=1.7.0'
  - <1.8.0
seeds:
  +schema: jaffle_raw
vars:
  truncate_timespan_to: '{{ current_timestamp() }}'

```

Note the preserved comments and formatting in the bog-standard parts of the YAML file.

### DIFF
--- a/dbt_meshify/storage/dbt_project_editors.py
+++ b/dbt_meshify/storage/dbt_project_editors.py
@@ -1,4 +1,6 @@
+import shutil
 from pathlib import Path
+from tracemalloc import start
 from typing import Dict, Optional, Set
 
 from dbt.contracts.graph.nodes import (
@@ -15,6 +17,7 @@ from loguru import logger
 
 from dbt_meshify.change import (
     ChangeSet,
+    DirectoryChange,
     EntityType,
     FileChange,
     Operation,
@@ -28,6 +31,20 @@ from dbt_meshify.utilities.contractor import Contractor
 from dbt_meshify.utilities.dependencies import DependenciesUpdater
 from dbt_meshify.utilities.grouper import ResourceGrouper
 from dbt_meshify.utilities.references import ReferenceUpdater
+
+
+def get_starter_project_path() -> Path:
+    """Obtain the path of a dbt starter project on the local filesystem."""
+
+    from importlib import resources
+
+    import dbt.include.starter_project
+
+    starter_path = Path(str(resources.files(dbt.include.starter_project)))
+    assert starter_path is not None
+    assert (starter_path / "dbt_project.yml").exists()
+
+    return starter_path
 
 
 class DbtSubprojectCreator:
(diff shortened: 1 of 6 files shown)


dbt_meshify/storage/dbt_project_editors.py
```diff
@@ -85,14 +103,29 @@ def _get_subproject_boundary_models(self) -> Set[str]:
...
+        starter_path: Path = get_starter_project_path() / "dbt_project.yml"
+        starter_dbt_project = YAMLFileManager.read_file(starter_path)
+
         contents = self.subproject.project.to_dict()
-        # was getting a weird serialization error from ruamel on this value
-        # it's been deprecated, so no reason to keep it
-        contents.pop("version")
+
         # this one appears in the project yml, but i don't think it should be written
         contents.pop("query-comment")
```
dave-connors-3, line 130: i remember this line being pretty /shrug at the time i wrote it -- any chance you tested removing this line and seeing if it worked? I feel like it may not in be in the yml from the starter project, and therefore not something we'd need to worry about anymore.

dave-connors-3 (approved):
🦞

dbt_meshify/storage/dbt_project_editors.py
nicholasyager: Based on my tests, if we remove it from here, it will be rendered in the sub-project's `dbt_project.yml`. Is that the desired behavior?

dave-connors-3: no i wouldn't say so 😅 i think my thought was that it may no longer be in the YML we start with now that we're using the official starter project file. if it's still there, I'd vote not to render it

Issue #210, opened 2024-05-29
Various issues returning `Exception: Unable to find a macro block with the name abcdef.`


b-per:
### Describe the bug

Tested with dbt-meshify==0.5.3 as 0.5.4 seems to be broken, at least on my machine.

Running `dbt-meshify split reference_data --select "group:xyz" -r` returns the error `Exception: Unable to find a macro block with the name abcdef.` when the macro abcdef is actually in the file

#### Use case 1

Create a macro with the following code

```
{% macro abcdef (
   param1 = '', 
   param2 = [],
  param3 = {}
) %}
 -- nothing to seee
{% endmacro %}
```

The regex [here](https://github.com/dbt-labs/dbt-meshify/blob/c6ae7673f2efb17c1ac2fe5f89d709f803e87aff/dbt_meshify/storage/jinja_blocks.py#L27) doesn't "find" the macro.

#### Use case 2

Create a generic test like this

```
{% test my_test(model,) %}
-- nothing interesting
{% endtest %}
```

The same piece of code searches the file for`block_type = "macro"` and `name = "test_my_test"`, so it raises an Exception as well.

### Suggestion

[Here](https://github.com/dbt-labs/dbt-meshify/blob/c6ae7673f2efb17c1ac2fe5f89d709f803e87aff/dbt_meshify/storage/jinja_blocks.py#L27), replace 
```
        for match in re.finditer(
            r"{%-?\s*" + block_type + r"\s*" + name + r"\s*([(a-zA-Z0-9=,_ \'\")]*)\s*-?%}",
            file_content,
            re.MULTILINE,
        ):
```

by 
```
        if block_type == "macro":
            block_type_new = "(macro|test)"
            name_new = f"({name}|{name[5:]})"
        else:
            block_type_new = block_type
            name_new = name

        pattern = r"{%-?[\s\n]+" + block_type_new + r"[\s\n]+" + name_new + r"([(a-zA-Z0-9=,_ \[\]{}'\s\n)]*)[\s\n]*-?%}" 
        for match in re.finditer(
            pattern,
            file_content,
            re.MULTILINE,
        ):
```

The handling of test/macro is ugly and should be done better, but at least, on the dbt project I am testing, it works!

PR #211, opened 2024-05-29
Fix for custom generic tests

dave-connors-3:
When someone creates a custom generic test, it gets rendered as a macro in the manifest. this means that when we compile all the macros in the project during startup of the group and split commands, we fail to find the correct test block. 

this PR
- updates the args to the `JinjaBlock.from_file` call to be `test` and trims `test_` (which dbt adds) to find the name of the custom generic test block when the macro path is under `tests/generic/`
- adds a custom generic test to the split project to ensure this logic works!

submitted in partial fulfillment of #210

### DIFF
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -340,8 +340,7 @@ class DbtProject(BaseDbtProject, PathedProject):
                 else ["tests"]
             ):
                 block_type = "test"
-                index = len(top_level_folder)
-                name = macro.name[index:]
+                name = macro.name[5:]
 
             blocks[unique_id] = JinjaBlock.from_file(
                 path=self.path / macro.original_file_path, block_type=block_type, name=name
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -330,11 +330,18 @@ class DbtProject(BaseDbtProject, PathedProject):
         for unique_id, macro in self.manifest.macros.items():
             block_type = "macro"
             name = macro.name
+            top_level_folder = macro.path.split("/")[0]
+
             if macro.package_name != self.name:
                 continue
-            if "tests/generic/" in macro.path:
+            if (
+                top_level_folder in self.project.test_paths
+                if self.project.test_paths
+                else ["tests"]
+            ):
                 block_type = "test"
-                name = macro.name[5:]
+                index = len(top_level_folder)
+                name = macro.name[index:]
 
             blocks[unique_id] = JinjaBlock.from_file(
                 path=self.path / macro.original_file_path, block_type=block_type, name=name
--- a/dbt_meshify/dbt_projects.py
+++ b/dbt_meshify/dbt_projects.py
@@ -140,7 +140,7 @@ class BaseDbtProject:
             return self._models
 
         self._models = {
-            node_name: node
+            node_name: node  # type: ignore
             for node_name, node in self.manifest.nodes.items()
             if node.resource_type == "model" and isinstance(node, ModelNode)
         }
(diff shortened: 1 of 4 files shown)


b-per (approved):

dbt_meshify/dbt_projects.py
```diff
@@ -328,11 +328,16 @@ def find_jinja_blocks(self) -> Dict[str, JinjaBlock]:
             )
 
         for unique_id, macro in self.manifest.macros.items():
+            block_type = "macro"
+            name = macro.name
             if macro.package_name != self.name:
                 continue
+            if "tests/generic/" in macro.path:
+                block_type = "test"
+                name = macro.name[5:]
```
nicholasyager, line 337: I don't feel super confident in using this hard-coded path. Would it make sense to use the test directory provided in `dbt_project.yml` here instead?

dbt_meshify/dbt_projects.py
```diff
@@ -140,7 +140,7 @@ def models(self) -> Dict[str, ModelNode]:
             return self._models
 
         self._models = {
-            node_name: node
+            node_name: node  # type: ignore
```
nicholasyager, line 143: 👍🏻

dbt_meshify/dbt_projects.py
```diff
@@ -328,11 +328,16 @@ def find_jinja_blocks(self) -> Dict[str, JinjaBlock]:
...
+            name = macro.name
             if macro.package_name != self.name:
                 continue
+            if "tests/generic/" in macro.path:
+                block_type = "test"
+                name = macro.name[5:]
 
             blocks[unique_id] = JinjaBlock.from_file(
-                path=self.path / macro.original_file_path, block_type="macro", name=macro.name
+                path=self.path / macro.original_file_path, block_type=block_type, name=name
```
nicholasyager, line 340: I really like that this function has been parameterized 👍🏻 🪨

nicholasyager (commented):
@dave-connors-3 The code in this PR makes sense for the most part. I had one small question about the use of `test/generic/` vs the `test-path` value in `dbt_project.yml`. Once that's sorted, we should be good to move forward.

dbt_meshify/dbt_projects.py
dave-connors-3: that makes sense to me! will update

dbt_meshify/dbt_projects.py
```diff
@@ -328,11 +328,23 @@ def find_jinja_blocks(self) -> Dict[str, JinjaBlock]:
...
+            block_type = "macro"
+            name = macro.name
+            top_level_folder = macro.path.split("/")[0]
+
             if macro.package_name != self.name:
                 continue
+            if (
+                top_level_folder in self.project.test_paths
+                if self.project.test_paths
+                else ["tests"]
```
nicholasyager, line 340: Beautiful logic 🪨

dbt_meshify/dbt_projects.py
```diff
@@ -328,11 +328,23 @@ def find_jinja_blocks(self) -> Dict[str, JinjaBlock]:
...
+            ):
+                block_type = "test"
+                index = len(top_level_folder)
```
nicholasyager, line 343: How is `index` being used?

nicholasyager (approved):
@dave-connors-3 One minor question about the `index` variable. Otherwise should be good to 🚢

dbt_meshify/dbt_projects.py
dave-connors-3: ah wait that's wrong -- the `5` should still be hardcoded to remove `test_`from the name of the macro to properly search the jinja for `my_test` and not `test_my_test`

nicholasyager: Yes! Thanks makes sense 👍🏻

PR #212, opened 2024-05-29
More jinja block escapes

dave-connors-3:
This PR solves the other half of beniots #210 !

This allows for list, dict, and newline characters in our jinja block search regex

I am sorry in advance, I branched from the other branch instead of main which i thought would be better but I see now is far worse. We should merge #211 first!


nicholasyager (approved):
@dave-connors-3 Thanks for adding these extra test cases and updating the Jinja regex accordingly. Great work!

nicholasyager (approved):

